PR #14, opened 2021-01-29
remove package.json requirement

lillianzhang331:
# Description

Remove package.json requirement to fix downstream buildpack issues for builds that need the Node runtime, but that may not have a package.json.

# Checklist:

- [x] Run these changes with `pack`
- [x] Make updates to `CHANGELOG.md`

### DIFF
--- a/buildpacks/nodejs/lib/detect.sh
+++ b/buildpacks/nodejs/lib/detect.sh
@@ -1,10 +1,5 @@
 #!/usr/bin/env bash
 
-detect_package_json() {
-	local build_dir=$1
-	[[ -f "$build_dir/package.json" ]]
-}
-
 write_to_build_plan() {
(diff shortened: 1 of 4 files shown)


buildpacks/nodejs/lib/detect.sh
```diff
@@ -1,10 +1,5 @@
 #!/usr/bin/env bash
 
-detect_package_json() {
-	local build_dir=$1
-	[[ -f "$build_dir/package.json" ]]
-}
-
```
danielleadams, line 7: We'll need to remove 

```
[[requires]]
node
```

from line 9 too!

danielleadams (changes requested):

danielleadams (approved):
🚀

PR #15, opened 2021-02-17
copy node modules dir path into env

lillianzhang331:
In order to make node modules available downstream, copy the path into the build ENV for subsequent buildpacks.

### DIFF
--- a/buildpacks/nodejs/lib/build.sh
+++ b/buildpacks/nodejs/lib/build.sh
@@ -167,7 +167,10 @@ set_node_modules_env() {
 
 	mkdir -p "${layer_dir}/env"
 	if [[ ! -s "${layer_dir}/env/NODE_MODULES_ENV" ]]; then
-		echo "$(cd "$(dirname "${layer_dir}/node_modules")"; pwd -P)/$(basename "${layer_dir}/node_modules")" >>"${layer_dir}/env/NODE_MODULES_ENV"
+		echo "$(
+			cd "$(dirname "${layer_dir}/node_modules")"
+			pwd -P
+		)/$(basename "${layer_dir}/node_modules")" >>"${layer_dir}/env/NODE_MODULES_ENV"
 	fi
 }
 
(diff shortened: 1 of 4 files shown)


buildpacks/nodejs/lib/build.sh
```diff
@@ -162,6 +162,15 @@ set_node_env() {
 	fi
 }
 
+set_node_modules_env() {
+	local layer_dir=$1
+
+	mkdir -p "${layer_dir}/env"
+	if [[ ! -s "${layer_dir}/env/NODE_MODULES_ENV" ]]; then
+		echo "$(cd "$(dirname "${layer_dir}/node_modules")"; pwd -P)/$(basename "${layer_dir}/node_modules")" >>"${layer_dir}/env/NODE_MODULES_ENV"
```
lillianzhang331, line 170: @danielleadams is this the right way to find path of node modules and copy it into NODE_MODULES_ENV?

buildpacks/nodejs/lib/build.sh
```diff
@@ -162,6 +162,18 @@ set_node_env() {
...
 
+set_node_modules_env() {
+	local layer_dir=$1
+
+	mkdir -p "${layer_dir}/env"
+	if [[ ! -s "${layer_dir}/env/NODE_MODULES_ENV" ]]; then
+		echo "$(
+			cd "$(dirname "${layer_dir}/node_modules")"
+			pwd -P
+		)/$(basename "${layer_dir}/node_modules")" >>"${layer_dir}/env/NODE_MODULES_ENV"
```
danielleadams, line 173: Can we rename the env var to `NODE_MODULES_PATH`? Otherwise it looks good!

danielleadams (approved):
🚀

buildpacks/nodejs/lib/build.sh
danielleadams: I think you may be able to do `echo "${layer_dir}/node_modules"`

PR #16, opened 2021-02-23
test meta buildpack with unreleased changes

danielleadams:
We want to test our metabuildpacks with the added changes of unreleased buildpacks while releasing with the released versions of buildpacks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,6 +35,7 @@ jobs:
           name: "Build and package buildpack"
           command: |
             package_toml="<< parameters.buildpack-dir >>/package.toml"
+
             if [[ -f "<< parameters.buildpack-dir >>/build.sh" ]]; then
               "./<< parameters.buildpack-dir >>/build.sh"
               package_toml="<< parameters.buildpack-dir >>/target/package.toml"
@@ -99,3 +100,4 @@ workflows:
                 - "buildpacks/npm"
                 - "buildpacks/typescript"
                 - "meta-buildpacks/nodejs"
+                - "test/meta-buildpacks/nodejs"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,7 +93,7 @@ workflows:
                 - "buildpacks/typescript"
               stack:
                 - "heroku-16"
-                - "heroku-18"
+                # - "heroku-18"
                 - "heroku-20"
 
       - package-buildpack:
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -32,15 +32,20 @@ jobs:
...
       - run:
-          name: "Build and package buildpack"
+          name: "Build and package buildpack with retries"
           command: |
-            package_toml="<< parameters.buildpack-dir >>/package.toml"
-            if [[ -f "<< parameters.buildpack-dir >>/build.sh" ]]; then
-              "./<< parameters.buildpack-dir >>/build.sh"
-              package_toml="<< parameters.buildpack-dir >>/target/package.toml"
-            fi
+            n=1
```
Malax, line 37: Is this to work around ECR download failures?

.github/scripts/release-workflow-package-push.sh
```diff
@@ -35,7 +36,7 @@ while IFS="" read -r -d "" buildpack_toml_path; do
 		echo "::set-output name=id::${buildpack_id}"
 		echo "::set-output name=version::${buildpack_version}"
 		echo "::set-output name=path::${buildpack_path}"
-		echo "::set-output name=address::${buildpack_docker_repository}@$(crane digest "${image_name}")"
+		echo "::set-output name=address::docker://${image_name}"
```
Malax, line 39: I had something similar before, it messed up things with `pack` and the registry. I ended up keeping the `address` as-is and added a `docker://` prefix where needed. Have you tried releasing after these changes?

.circleci/config.yml
danielleadams: yes, I made a similar change here: https://github.com/heroku/pack-images/pull/156

.github/scripts/release-workflow-package-push.sh
danielleadams: I haven't tested yet, but I will change it and test it

Malax: Can I suggest diffing with https://github.com/heroku/buildpacks-jvm/commit/14889c54e0a0b598d400a6126adc95fd937e33ec? It looks like some of the code here is based on an older version that I since then fixed. I know for sure that the changes work in the commit above.

Malax (approved):

.github/scripts/release-workflow-package-push.sh
danielleadams: done 👍🏼

PR #28, opened 2021-03-02
Flush cache on stack change

lillianzhang331:
If the stack changes, the cache should be cleared. This is currently not happening. The stack should be stored in a persisted build file (store.toml) so that it can be read on the next run.


buildpacks/nodejs/bin/build
```diff
@@ -20,6 +20,11 @@ info "Node.js Buildpack"
 
 rm -rf node_modules
 
+install_or_reuse_toolbox "$layers_dir/toolbox"
```
lillianzhang331, line 23: toolbox needs to be installed first so that yj can be used to parse store.toml file

buildpacks/nodejs/lib/build.sh
```diff
@@ -16,6 +16,28 @@ source "$bp_dir/lib/utils/log.sh"
...
+	local layers_dir=$1
+
+	if [[ ! -f "${layers_dir}/store.toml" ]]; then
+		touch "${layers_dir}/store.toml"
+		cat <<TOML >"${layers_dir}/store.toml"
+[metadata]
+last_stack = "$CNB_STACK_ID"
+TOML
+	else
+		if [[ $CNB_STACK_ID != $(toml_get_key_from_metadata "${layers_dir}/store.toml" "last_stack") ]]; then
```
lillianzhang331, line 29: `$(cat "${layers_dir}/store.toml" | grep "last_stack")` outputs `last_stack` and not `last_stack = "heroku-18"` like its supposed to

buildpacks/nodejs/bin/build
```diff
@@ -20,6 +20,11 @@ info "Node.js Buildpack"
...
+export PATH=$layers_dir/toolbox/bin:$PATH
```
danielleadams, line 24: I think we can delete line 23 and 24 now

danielleadams (approved):
looks good!

PR #30, opened 2021-03-04
fail if two lock files are detected

lillianzhang331:
When the build starts, if there are two lock files present, the build should fail immediately. This is done in the npm and yarn buildpacks.


danielleadams (approved):

danielleadams:
Tested this with pack locally, we'll need to add testing for errors and error exit codes later.

PR #50, opened 2021-05-06
Fix nodejs-function dependency versions

Malax:
Fixes for #48:

Those were using the versions of the next release, most likely copied from the test meta-buildpack. I also updated the CircleCI config to package this buildpack which would have prevented this issue.


Malax:
CI is currently failing (and stuck in an endless-loop), packaging works fine locally. Investigating.

edmorley (approved):
Thank you!

PR #47, opened 2021-05-05
Add nodejs-function-invoker buildpack

edmorley:
This replaces the various Riff-based invoker buildpacks, now that Riff has been sunset. This buildpack's purpose is to vendor the new Node.js functions invoker:
https://github.com/forcedotcom/sf-fx-runtime-nodejs

For Riff invoker buildpack prior art, see:
https://github.com/forcedotcom/nodejs-sf-fx-buildpack
https://github.com/projectriff/node-function-buildpack

Also takes inspiration from the JVM invoker buildpack:
https://github.com/heroku/buildpacks-jvm/tree/main/buildpacks/jvm-function-invoker
...though with a number of changes since this repo has a different style of buildpack structure (to allow for unit testing), along with different buildpack utility functions (eg for logging), and uses shpec rather than Rapier for testing.

This initial version of the buildpack doesn't implement caching, since whilst the invoker is under active development, the release assets will be regularly changing, so using unversioned URLs and frequently changing archive SHA.

This buildpack also currently doesn't pass `$PORT` to the entrypoint, since the invoker currently doesn't support it (and may read directly from the env anyway, when there is support for custom ports).

The Circle CI shellcheck version has been bumped to pick up a fix that reduces the number of `source=` annotations required to prevent `SC1090` errors.

After this PR is merged and the buildpack published, the meta-buildpack `nodejs-function` will be updated in another PR (#48), to reference this buildpack instead of the Riff-based buildpacks.

Closes GUS-W-9115013.

PR #52, opened 2021-05-06
Run initial checks before build

danielleadams:
We have a few checks that should be run before a build attempts to start in the function invoker buildpack. The first one should check if the file that is listed as "main" in the function's package.json should be present. If the file is not present, the build will fail.


buildpacks/nodejs-function-invoker/fixtures/package-json-no-main/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "main": "index.js"
```
edmorley, line 2: This fixture is named `package-json-no-main` which made me think it would contain a `package.json` that didn't have a `main` key, but it's actually that the `main` file is missing. Do we need two fixtures here - one with a missing main key named perhaps `package-json-no-main`, and a second with `index.js` missing (like this fixture) named something else?

Also, I realise now I missed out the `main` entry (and index.js) from my fixtures added in #47 - would you mind adding it to them now? These could be used to test the "working" case of `run_initial_checks`.

buildpacks/nodejs-function-invoker/lib/failures.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+fail_on_no_main_file() {
+	build_dir=$1
+	fn_entry_file=$(json_get_key "$build_dir/package.json" ".main")
```
edmorley, line 5: Should this also handle the case where the `main` key doesn't exist at all in `package.json`?

buildpacks/nodejs-function-invoker/shpec/build_shpec.sh
```diff
@@ -19,6 +19,20 @@ create_temp_app_dir() {
 }
 
 describe "lib/build.sh"
+	describe "run_initial_checks"
```
edmorley, line 22: I think it would be worth also testing the "working" case, plus also the "package.json doesn't have a 'main' key" case.

danielleadams:
@edmorley thanks for the comments! Converted to done - I need to add the other use case you pointed out (no "main" key in package.json) and I'm having a hard time importing the fixtures for the unit tests, so I need to figure out what is going on there.

buildpacks/nodejs-function-invoker/fixtures/package-json-no-main/package.json
danielleadams: Agreed, I can make `package-json-no-main` and `package-json-no-main-file` or something? I'll add to the fixtures too.

buildpacks/nodejs-function-invoker/lib/failures.sh
danielleadams: Yeah good call, I'll create another failure method for this.

danielleadams:
@edmorley this is ready for re-review if you are around!

buildpacks/nodejs-function-invoker/bin/build
```diff
@@ -4,8 +4,11 @@
...
 
 layers_dir="${1:?}"
+build_dir=$(pwd)
 
 source "${CNB_BUILDPACK_DIR}/lib/build.sh"
 
+run_initial_checks "$build_dir"
+
 install_runtime "${layers_dir}"
 write_launch_toml "${layers_dir}" "$(pwd)"
```
edmorley, line 14:
```suggestion
write_launch_toml "${layers_dir}" "${build_dir}"
```

buildpacks/nodejs-function-invoker/lib/failures.sh
```diff
@@ -0,0 +1,21 @@
...
+		exit 1
+	fi
+}
+
+fail_on_no_main_key() {
+	build_dir=$1
+	fn_entry_file=$(json_get_key "$build_dir/package.json" ".main")
+
+	if [[ -z $fn_entry_file ]]; then
+		error "Key at \"main\" in package.json is missing. Make sure to use \"main\" to specify your root function file."
```
edmorley, line 18:
```suggestion
		error "Key \"main\" in package.json is missing. Make sure to use \"main\" to specify your root function file."
```

edmorley (approved):
Looks great! :-)

PR #58, opened 2021-05-11
Detect for type=function in project.toml

hone:
See https://salesforce.quip.com/tLL9AiScqg5q for the schema (Salesforce internal link). Sadly, I don't have a public link atm.


edmorley (approved):
Thank you for this!

Longer term I think we need to improve the UX for detection failures (eg add appropriate error messages, and also look at pack/lifecycle to see how this works in the specified buildpack vs no-specified-buildpacks cases), but for now this is great :-)

Could you rebase this on `main` and adjust the CHANGELOG addition accordingly?

hone:
@edmorley sorry for the  delay. I've rebased and pushed up those changes.

edmorley (approved):

PR #75, opened 2021-05-27
remove heroku-16 test from CI

danielleadams:


edmorley (approved):
Ah didn't know this repo tested on Heroku-16 -- since pack-images doesn't have a Heroku-16 based image, I thought none of the CNBs would, but I suppose this is for unit tests using shpec :-)

PR #102, opened 2021-07-28
Add docs on releasing buildpacks

schneems:


Malax (commented):
Nice to see this being documented, thanks! Does it make sense to move this documentation to the languages-team repository? The Java repository works the same way and duplicating the docs seems terrible. We can still link to them in here I think. WDYT?

Malax:
My previous comment got two 👍🏻 , do you want to change this PR so that it points to a version of this in our `languages-team` repo @schneems?

schneems:
One of those thumbs was mine 😄 . I'm in the process of converting the docs. It's not a straight copy/pasta since this references node specific URLs. One downside of moving this over to the language playbooks is that it makes the process less open. Other teams looking to adopt and publish buildpacks while they're in this early stage of development might not have access. I don't think it's enough to be a blocker, but it's a consideration. 

As CNB authors we have a vested interest in the long-term success of the ecosystem so we don't end up with another V2 legacy system no one else adopts or uses. I want us on the lookout for ways we can support that community in the open when it doesn't largely conflict with our own needs to ship products.

schneems:
Merged in https://github.com/heroku/languages-team/pull/74. Updated to point at the internal doc.

Malax (approved):

Malax:
@schneems I don't think we'll keep the current process when the migration to Rust is finished. It's quite complex because it has to support shimming V2 buildpacks, introduces a layer of build abstraction to support bash AND Rust buildpacks, etc. 

Ideally, the whole build and packaging process will be handled by Cargo mid-term.

PR #108, opened 2021-08-03
Improve shfmt complaints

noroutine:
@schneems, this should do the git diff and exit properly


edmorley:
I've enabled Circle CI on forks, so CI will run automatically now :-) (We tend to have this turned off on many of the other repos, since CI there needs secrets so would fail anyway, however this repo doesn't need secrets to run CI)

noroutine:
Thank you @edmorley 

Where should I put changelog line about this PR? It doesn't really fall into any of existing ones

joshwlewis (approved):

schneems:
Awesome, thanks!

PR #132, opened 2021-10-04
Trim whitespace on release build pack

schneems:
Currently if you attempt to release a buildpack with id `heroku/nodejs   ` (with whitespace before or after) it will fail. It also takes a few minutes to fail and isn't 100% obvious in the message that the failure is due to whitespace.

This trims whitespace from the input on both release scripts. I also modified the error message to quote the variable so if it's used in the future without being trimmed first, the error would be easier to detect.

Other considerations:

- Could we trim the input earlier? Ideally not having to do this in every script would be ideal.
- It takes a long time running on actions before this will execute, we should check the input build pack id early and fail early to give fast feedback in the future.
- These scripts are not tested and should be modified with extreme caution. We should extract as much functionality into bash functions as possible and write some tests against them.
- Similar to above: Ideally we would lint these changes.

### DIFF
--- a/.github/scripts/release-workflow-package-push.sh
+++ b/.github/scripts/release-workflow-package-push.sh
@@ -6,6 +6,7 @@ shopt -s globstar
 # It requires the following environment variables to function correctly:
 #
 # REQUESTED_BUILDPACK_ID - The ID of the buildpack to package and push to the container registry.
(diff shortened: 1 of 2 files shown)


edmorley (approved):
Thank you :-)

PR #159, opened 2021-11-08
Fix yarn install and upgrade to Buildpack API 0.6

jkutner:
Buildpack API 0.6 requires the `launch`, `cache`, and `build` keys to be under the `[types]` table.


joshwlewis (approved):
Looks good, thanks. Needs a changelog entry, though!

PR #163, opened 2021-11-10
Post-release updates: heroku/nodejs-yarn 0.1.7

joshwlewis:
We were able to release the image for heroku/nodejs-yarn 0.1.7 here: https://github.com/buildpacks/registry-index/issues/2147

However, the rest of the release script failed (including generating this PR) [here](https://github.com/heroku/buildpacks-nodejs/actions/runs/1445418777). I'm creating this PR manually, because the release action won't ever pass for the same release again (the buildpack registry rejects duplicates).

### DIFF
--- a/buildpacks/yarn/CHANGELOG.md
+++ b/buildpacks/yarn/CHANGELOG.md
@@ -4,6 +4,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


joshwlewis:
I've yanked 0.1.7 from the registry: https://github.com/buildpacks/registry-index/issues/2151.

I've also deleted the tag here on GitHub.

We never published this buildpack to ECR, so it's like this release (0.1.7) never happened.

PR #178, opened 2022-01-18
Drop the typescript buildpack

joshwlewis:
As discussed in [DR: TypeScript CNB Future](https://salesforce.quip.com/hEJCAHKAJhAr), TypeScript functionality is well covered by `heroku/nodejs-npm` and `heroku/nodejs-yarn` buildpacks, and we don't need a dedicated CNB for it.

For example, this `package.json` will force the `heroku/nodejs` buidpack to install and compile TypeScript:

```
{
  "scripts": {
    "build": "tsc",
  },
  "devDependencies": {
    "typescript": "4.5.4",
    ...
  }
}
```

GUS-W-10455655.


meta-buildpacks/nodejs-function/buildpack.toml
```diff
@@ -19,11 +19,6 @@ version = "0.7.4"
 id = "heroku/nodejs-npm"
 version = "0.4.4"
```
edmorley, line 20: Since this is a technically a breaking change (eg removal of `CUSTOM_TSCONFIG` support, even if it's not really an issue in practice) it seems this needs a major version bump (since the automation has pre-emptively bumped the versions, but only for a patch release).

In addition to this file `meta-buildpacks/nodejs/buildpack.toml` will also need updating (but the affected lines are out of the range on which GitHub will let me comment).

```suggestion
version = "0.5.0"
```

edmorley (approved):

meta-buildpacks/nodejs-function/buildpack.toml
joshwlewis: Your suggested change seems to be bumping `heroku/nodejs-npm` version. Did you mean to bump `heroku/nodejs-function` instead? The latter would make sense to me. The meta-buildpacks have a signifcant change warranting a major (or minor, since we're pre 1.0) bump here, but the `heroku/nodejs-npm` buildpack doesn't have any functionality changes.

joshwlewis: Done in a7db428f099433c850e3a7fd000618c3e0c21deb.

edmorley: Oh sorry I picked the wrong line, yeah what you said :-)

maxgio92:
Hello @joshwlewis, as so, is the Typescript build (`tsc`) supposed to be run at runtime by default, right?

joshwlewis:
@maxgio92, no, `tsc` is expected to run at build time. If you have `tsc` or similar in your package.json's `scripts.build`, we'll run it after dependencies are installed.

maxgio92:
Thank you @joshwlewis . So, with your CNB no need to put the `tsc` run as `scripts.prestart`, right?

joshwlewis:
Yes, I'd suggest running `tsc` during `scripts.build` instead, that way your compiled JS is part of the built image.

maxgio92:
Totally agree, thank you @joshwlewis

PR #186, opened 2022-01-27
Fix `tsc: not found` for typescript rebuilds

joshwlewis:
## Issue

In the scenario where a TypeScript app build has the correct Node.js version cached, but something in the `package.json` or `package-lock.json` has changed causing the `node_modules` cache to be invalid, that build would fail with `tsc: not found`. 

## Why it was happening

On a build without any cache, TypeScript builds would pass. But this may have been accidental. We were creating a `env.build/NODE_ENV.override` file, but we'd unintentionally delete it a short time later with a `rm -rf` during `install_or_reuse_node()`. Since `NODE_ENV` was unset, that meant that `npm ci` and/or `npm install` would install all `devDependencies`, including `typescript`.

On a build with a valid `nodejs` cache, the `rm -rf` mentioned never happens. So, for these builds `NODE_ENV` is set to `production` during `npm ci` or `npm install`, and `devDependencies` are NOT installed, so any `tsc` calls fail.

On a build with a valid `nodejs` and `node_modules` cache, the `rm -rf` still doesn't happen, but it doesn't matter because we don't run `npm install`, we instead restore the module cache, which includes `tsc` from the previous build. 

## How it's getting fixed

- We make sure that the `NODE_ENV` file is created AFTER the `nodejs` layer is cleaned out. 
- since we were also setting `env.launch/NODE_ENV.override`, consolidated these functions/files to create just one `env/NODE_ENV.override`that should serve both purposes.
- Now that `NODE_ENV` is always `production` during builds, `npm install` and `npm ci` will both skip `devDependencies`. Adding `--production=false` to these calls ensure we still install `devDependencies` like `typescript`.

GUS-W-10034673

### DIFF
--- a/buildpacks/nodejs-engine/CHANGELOG.md
+++ b/buildpacks/nodejs-engine/CHANGELOG.md
@@ -3,6 +3,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Ensure NODE_ENV is set consistently during build, no matter the cache state ([186](https://github.com/heroku/buildpacks-nodejs/pull/186)
 
 ## [0.7.4] 2021/06/15
 - Change node engine version from 12 to 14 ([#40](https://github.com/heroku/buildpacks-node/pull/40))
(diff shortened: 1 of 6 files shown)


edmorley (approved):
(completely a rubber stamp :-))

francescopersico:
@joshwlewis i think there is the same problem when using yarn. From my test devDependencies get not installed so tsc is not found.

PR #201, opened 2022-03-03
Let yarn install install also devDependencies

francescopersico:
We have a similar problem described here https://github.com/heroku/buildpacks-nodejs/pull/186 but when using **yarn**.

### DIFF
--- a/buildpacks/yarn/lib/build.sh
+++ b/buildpacks/yarn/lib/build.sh
@@ -59,10 +59,10 @@ install_modules() {
 
 	if detect_yarn_lock "$build_dir"; then
 		echo "---> Installing node modules from ./yarn.lock"
-		yarn install
+		yarn install --production=false
 	else
 		echo "---> Installing node modules"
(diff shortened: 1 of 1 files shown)


francescopersico:
> I think this will work for now. Longer term, we need a way to support folks who don't want to install `devDependencies` at all. We do this in the classic buildpack by looking at what `NODE_ENV` is set to.

@joshwlewis i agree with you about the longer term solution. Meanwhile do you think we can merge this PR and release a new version of the pack so we can also use it for typescript projects based on yarn?

schabr06:
This commit has broken the builds that are using Yarn2+ due to the --production=false flag being added on the `yarn install` command on lines 62 and 65 in the [buildpacks/yarn/lib/build.sh](https://github.com/heroku/buildpacks-nodejs/pull/201/files#diff-3e1bc704c9ceca5dee10b3672b975b57e461753ef8bcc0e15daa23512806713f)

PR #206, opened 2022-03-09
Install musl tools, rust, and libcnb-cargo prior to release

joshwlewis:
The release buildpack script is failing for rust based buildpacks: https://github.com/heroku/buildpacks-nodejs/actions/runs/1959345733.

We need to install musl-tools, rust, and libcnb-cargo so that we can build/release the libcnb-based buildpacks like `heroku/nodejs-engine` that was introduced in #184.

This code was lifted from https://github.com/heroku/buildpacks-jvm/blob/main/.github/workflows/release.yml#L31-L44

GUS-W-10729118

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,6 +28,20 @@ jobs:
           password: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         env:
           AWS_REGION: us-east-1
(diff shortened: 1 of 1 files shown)


colincasey (approved):

joshwlewis:
It worked: https://github.com/heroku/buildpacks-nodejs/actions/runs/1959502868

PR #214, opened 2022-03-23
Set package.json scripts.start as a default process

### LINKED ISSUE #211: Node.js app's "start" script not used anymore
Since #184, it looks like a node.js project needs a Procfile so that the buildpack finds its launch command.

Since about 3 hours ago (coincidentally, as this was merged: https://github.com/heroku/pack-images/pull/231), our automated tests started failing. These tests use a very basic, no dependencies, `package.json` that looks like this:

```json
{
    "name": "hello-node",
    "version": "1.0.0",
    "main": "index.js",
    "scripts": {
        "start": "node index.js"
    }
}
```

Ever since this change, we started seeing:

```
ERROR: failed to launch: determine start command: when there is no default process a command is required
```

We've resolved this by adding a Procfile like:

```procfile
web: npm start
```

Is this change intentional?

joshwlewis:
In #184, we updated the buildpack api version for `heroku/npm` to `0.6`. This had the unintended side effect of the `web` process no longer being flagged as a default process. In earlier buildpack API versions, overwriting a default process with an undefined default would result in a process flagged as default. Newer buildpack API versions will prefer the default setting of the overwriting process.

This PR ensures the `web` process affiliated with `package.json`'s `scripts.start` is flagged as a default process, for both `yarn` and `npm`. 

Fixes #211 

[W-10888625](https://gus.my.salesforce.com/one/one.app#/alohaRedirect/a07EE00000tVabFYAS)

### DIFF
--- a/buildpacks/npm/CHANGELOG.md
+++ b/buildpacks/npm/CHANGELOG.md
@@ -4,6 +4,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- The `web` process affiliated with `package.json`'s `scripts.start` is now a `default` process ([#214](https://github.com/heroku/buildpacks-nodejs/pull/214))
+
 ## [0.5.0] 2022/03/09
 
 - Upgraded to buildpack api 0.6 ([#184](https://github.com/heroku/buildpacks-nodejs/pull/184))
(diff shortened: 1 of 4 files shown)


edmorley (approved):
(I presume you'll add an integration test for this when you write the libcnb-tests for this repo, once the remaining conversions are performed? :-))

PR #215, opened 2022-03-23
package.json version now optional

### LINKED ISSUE #212: version is now required in package.json
This appears to be a regression (or undocumented change) from #184 

Projects without a `version` defined in `package.json` will fail to build with the following error:

```
  ===> BUILDING
  
  [Heroku Node.js Engine Buildpack]
  
  [Checking Node.js version]
  
  [Error: Node.js engine package.json error]
  Couldn't parse package.json: Could not parse package.json: missing field `version` at line 78 column 1
  ERROR: failed to build: exit status 42
  ERROR: failed to build: executing lifecycle: failed with status code: 145
```

joshwlewis:
Fixes #212

### DIFF
--- a/buildpacks/nodejs-engine/CHANGELOG.md
+++ b/buildpacks/nodejs-engine/CHANGELOG.md
@@ -4,6 +4,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


edmorley (approved):
Is it worth changing one/some of the integration test fixtures so that they omit all fields from `package.json` aside from `name` and `engines`?

joshwlewis:
> Is it worth changing one/some of the integration test fixtures so that they omit all fields from package.json aside from name and engines?

Done in e08e917.

PR #221, opened 2022-03-23
make etag optional part of the inventory

hone:
To let others create an inventory.toml, I'd like to make `etag` optional. From what I can tell this is just an artifact of nodebin using S3. It's not being used anywhere in the buildpack. I wonder if we should even remove this.


hone:
I'm not sure there is an applicable CHANGELOG entry here?

joshwlewis (approved):

PR #224, opened 2022-03-29
Convert heroku/nodejs-function-invoker to libcnb.rs implementation

joshwlewis:
This changes `heroku/nodejs-function-invoker` from a bash based buildpack to a rust and libcnb.rs version. 

Most of the functionality stays the same, but there are some minor differences:

- `/opt/run.sh` has been removed from the source code and is no longer installed in the resulting image. It really wasn't needed as we're now using a bash launch process instead.
- `yj` is no longer installed during `detect`, and is no longer required during `build`. Previously, this buildpack expected `yj` to be installed (by the `heroku/nodejs-npm` buildpack).

[W-10915212](https://gus.my.salesforce.com/one/one.app#/alohaRedirect/a07EE00000tsBWkYAM)


buildpacks/nodejs-function-invoker/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+                            "--workers",
+                            "2",
+                            "--host",
+                            "::",
+                            "--port",
+                            "${PORT:-8080}",
+                            "--debug-port",
+                            "${DEBUG_PORT:-}",
+                        ])
+                        .default(true)
```
Malax, line 93: Nit: Even though it is the default, we could add `.direct(false)` here to explicitly show we expect this to be run via `bash`.

buildpacks/nodejs-function-invoker/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+                    }
+                }
+            }
+            err => {
+                log_error("Internal Buildpack Error", err.to_string());
+                100
+            }
+        }
+    }
+}
```
Malax, line 125: This could leverage `libherokubuildpack::on_error_heroku` to simplify error handling a little bit and benefit from cross-language error handling that will be eventually added to `libherokubuildpack`.

buildpacks/nodejs-function-invoker/src/main.rs
joshwlewis: Fixed in https://github.com/heroku/buildpacks-nodejs/pull/224/commits/5e1b6e50bcdb49e58c7f963ef8f43883b142f1e0

buildpacks/nodejs-function-invoker/src/function.rs
```diff
@@ -0,0 +1,132 @@
...
+use libherokubuildpack::toml_select_value;
+use std::path::PathBuf;
+use thiserror::Error;
+use toml::Value;
+
+pub fn is_function<P>(d: P) -> bool
+where
+    P: Into<PathBuf>,
+{
+    let dir: PathBuf = d.into();
```
Malax, line 12: Ultra-nit: the explicit type can be removed since `Into<PathBuf>` is the only `Into` available for `P`.

buildpacks/nodejs-function-invoker/src/layers/runtime.rs
```diff
@@ -0,0 +1,106 @@
...
+            .map_err(RuntimeLayerError::NpmCommandError)
+            .and_then(|output| {
+                if output.status.success() {
+                    Ok(output)
+                } else {
+                    io::stdout().write_all(&output.stdout).ok();
+                    io::stderr().write_all(&output.stderr).ok();
+                    Err(RuntimeLayerError::NpmInstallError(output.status))
+                }
+            })?;
```
Malax, line 76: Nit: While this works, explicitly writing stdout/err can be omitted if we use `spawn` instead of `output` which will pipe stdout and stderr to the parent process (this buildpack).

```suggestion
            .spawn()
            .and_then(|mut child| child.wait())
            .map_err(RuntimeLayerError::NpmCommandError)
            .and_then(|status| {
                status.success()
                    .then(|| ())
                    .ok_or( RuntimeLayerError::NpmInstallError(status))
            })?;
```

joshwlewis: I think this would always send stdout and stderr back to the build log? These logs are noisy and uninteresting in the happy path, so I'm trying to suppress them for successful installs.

buildpacks/nodejs-function-invoker/src/main.rs
joshwlewis: I didn't use `on_error_heroku` in `heroku/nodejs_engine` mostly because I didn't see the value. I still don't see much value, but I guess it might be there one day. I don't feel very strongly, so it's fixed in https://github.com/heroku/buildpacks-nodejs/pull/224/commits/8040d4177082b4da311f033245da498fd7301dfe.

buildpacks/nodejs-function-invoker/src/layers/runtime.rs
Malax: Ah, I totally missed that detail, oops. Nevermind then, sorry :) Might be worth a comment in the source?

joshwlewis: Added a comment and also dropped the `if {} else {}` in favor of a `.then()`, which I think may have been an implicit gripe. =P

colincasey (approved):

PR #231, opened 2022-04-04
Update heroku/procfile to 1.0.0

joshwlewis:
The new libcnb.rs `heroku/procfile` is ready to go. Let's use it!

[W-10475648](https://gus.my.salesforce.com/one/one.app#/alohaRedirect/a07EE00000hwHhMYAU)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
       - checkout
       - setup_remote_docker
       - pack/install-pack:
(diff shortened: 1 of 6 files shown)


edmorley (approved):
Since the new procfile buildpack switches to using the wildcard stack, you might be affected by:
https://github.com/buildpacks/pack/pull/1406

If so, updating to Pack 0.24.1 will resolve it :-)

PR #237, opened 2022-04-05
Fix inventory automation when there are no changes

joshwlewis:
In #235, there were no new node versions, but we opened a PR anyway. The change (added `-`) to the changelog triggered a PR to be opened.

This PR uses `xargs` to avoid touching the changelog when there is no diff message.

[W-10942589](https://gus.my.salesforce.com/one/one.app#/alohaRedirect/a07EE00000uDs1SYAS)


Malax (approved):
🎉

PR #250, opened 2022-04-12
Implement heroku/nodejs-yarn in libcnb.rs

joshwlewis:
This is a complete rewrite of heroku/nodejs-yarn. This PR converts the buildpack from Bash to Rust and leverages github.com/heroku/libcnb.rs (like we already do with heroku/nodejs-engine and heroku/nodejs-function-invoker).


buildpacks/nodejs-yarn/src/layers/dist.rs
```diff
@@ -0,0 +1,96 @@
...
+#[derive(Deserialize, Serialize, Clone, PartialEq, Eq)]
+pub struct DistLayerMetadata {
+    layer_version: String,
+    yarn_version: String,
+    stack_id: StackId,
+}
+
+#[derive(Error, Debug)]
+pub enum DistLayerError {
+    #[error("Couldn't create tempfile for Node.js distribution: {0}")]
```
colincasey, line 28: should this and the following errors reference `Yarn` instead of `Node.js`?

buildpacks/nodejs-yarn/src/main.rs
```diff
@@ -0,0 +1,156 @@
...
+            .join("yarn.lock")
+            .exists()
+            .then(|| {
+                DetectResultBuilder::pass()
+                    .build_plan(
+                        BuildPlanBuilder::new()
+                            .provides("yarn")
+                            .requires("yarn")
+                            .provides("node_modules")
+                            .requires("node_modules")
```
colincasey, line 44: 2 questions here:
* what does the `node_modules` buildpack do?  
* would the PnP feature from Yarn 2 require this?

joshwlewis: These aren't buildpacks per se. This is basically a list of things that this buildpack provides and requires. These strings are used to evaluate whether a group of buildpacks are compatible and should pass detection. 

This buildpack requires `node`, so it won't pass detection unless another buildpack in the group provides `node` (which heroku/nodejs-engine does). 

This buildpack provides `node_modules`, which would allow any other buildpack to require `node_modules`, which is a way to say "My buildpack expects another buildpack to install node_modules". It's also requiring `node_modules` so that it will pass detection, even if no other buildpack in the group is asking for `node_modules`.

I mostly chose to go with `node_modules` because 1) that's what the bash buildpack was doing 2) the paketo buildpacks also use `node_modules` and 3) I kind of thought the distinction between a pnp.js and a `node_modules` folder weren't relevant for downstream buildpacks -- they both provide modules for node to use.

buildpacks/nodejs-yarn/src/layers/dist.rs
joshwlewis: Hah, yeah. I'm still iterating pretty heavily on this.. lots of WIP =P

buildpacks/nodejs-yarn/src/layers/deps.rs
```diff
@@ -90,14 +90,19 @@ impl Layer for DepsLayer {
...
-            args.append(&mut vec!["--production", "false"]);
+            args.append(&mut vec![
+                "--cache-folder",
+                &cache_folder,
+                "--frozen-lockfile",
+                "--production",
+                "false",
+            ]);
+        } else {
+            args.append(&mut vec!["--immutable", "--immutable-cache"])
```
colincasey, line 105: @joshwlewis the [--inline-builds option](https://yarnpkg.com/cli/install#options-inline-builds) might be useful to add here too

buildpacks/nodejs-yarn/src/layers/deps.rs
```diff
@@ -90,14 +90,19 @@ impl Layer for DepsLayer {
...
-            args.append(&mut vec!["--cache-folder", &cache_folder]);
-        }
         if self.yarn_major_version == "1" {
-            args.append(&mut vec!["--production", "false"]);
+            args.append(&mut vec![
+                "--cache-folder",
+                &cache_folder,
+                "--frozen-lockfile",
+                "--production",
+                "false",
```
colincasey, line 102: Yarn v1 expects either `--production` or `--production=true|false` according to [the docs](https://classic.yarnpkg.com/en/docs/cli/install#toc-yarn-install-production-true-false)

buildpacks/nodejs-yarn/src/layers/deps.rs
```diff
@@ -0,0 +1,100 @@
...
+/// unbound cache growth.
+pub(crate) struct DepsLayer {
+    pub(crate) yarn: Yarn,
+}
+
+#[derive(Deserialize, Serialize, Clone, PartialEq)]
+pub(crate) struct DepsLayerMetadata {
+    // Using float here due to [an issue with lifecycle's handling of integers](https://github.com/buildpacks/lifecycle/issues/884)
+    cache_usage_count: f32,
+    layer_version: String,
```
schneems, line 26: Seems like a good idea to futureproof metadata. I wonder if we should do more to promote this somehow (in the future).

schneems (commented):
Mostly looks good. One failing test:


```
       +[detector] ======== Error: heroku/nodejs-yarn@0.3.0 ========
       +[detector] fork/exec /cnb/buildpacks/heroku_nodejs-yarn/0.3.0/bin/detect: no such file or directory
       +[detector] err:  heroku/nodejs-yarn@0.3.0
       +[detector] 2 of 3 buildpacks participating
       +[detector] heroku/nodejs-engine 0.8.13
       +[detector] heroku/nodejs-npm    0.5.3
```

buildpacks/nodejs-yarn/src/layers/deps.rs
joshwlewis: Yeah, I've done this on three buildpacks now. My thinking is that anytime I know I'm going to make a breaking change to the cache or metadata, I can bump this.

buildpacks/nodejs-yarn/src/cmd.rs
```diff
@@ -0,0 +1,120 @@
...
+) -> Result<(), Error> {
+    let mut args = vec!["install"];
+    if yarn_line == &Yarn::Yarn1 {
+        args.push("--production=false");
+        args.push("--frozen-lockfile");
+    } else {
+        args.push("--immutable");
+        args.push("--inline-builds");
+        if zero_install {
+            args.push("--check-cache");
```
colincasey, line 93: won't this cause yarn to re-download the packages?  

https://yarnpkg.com/cli/install#options-check-cache

buildpacks/nodejs-yarn/src/cmd.rs
```diff
@@ -0,0 +1,120 @@
...
+    zero_install: bool,
+    yarn_env: &Env,
+) -> Result<(), Error> {
+    let mut args = vec!["install"];
+    if yarn_line == &Yarn::Yarn1 {
+        args.push("--production=false");
+        args.push("--frozen-lockfile");
+    } else {
+        args.push("--immutable");
+        args.push("--inline-builds");
```
colincasey, line 91: very glad you're including this flag :)

buildpacks/nodejs-yarn/src/cmd.rs
joshwlewis: Hmm, yeah it will. I think this is more appropriate for CI than for prod builds. I'll drop it.

buildpacks/nodejs-yarn/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+            .app_dir
+            .join("yarn.lock")
+            .exists()
+            .then(|| {
+                DetectResultBuilder::pass()
+                    .build_plan(
+                        BuildPlanBuilder::new()
+                            .provides("yarn")
+                            .requires("yarn")
+                            .provides("node_modules")
```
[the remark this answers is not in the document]
joshwlewis: I was thinking of `node_modules` here as a concept, not the actual folder. As it is downstream buildpacks could require `node_modules` and that should mean modules are installed, but the downstream buildpack shouldn't have to know whether `node_modules` or `.pnp.js` was the result. Regardless of the mode, dependencies are installed and commands like `yarn test` should work downstream.

colincasey (approved):

edmorley (commented):
I didn't look too closely at the buildpack implementation (I defer to others there :-)) - but I noticed that the yarn buildpack currently isn't tested in the `package-buildpack` CI job - should it be?
https://github.com/heroku/buildpacks-nodejs/blob/3b3cfc1cd6c4b3d4df063812bb2d24363eabf627/.github/workflows/ci.yml#L97-L108

Malax (changes requested):
Preliminary review to get out what I have before EOD. Looking good overall! 👍🏻

common/nodejs-utils/src/package_json.rs
```diff
@@ -6,24 +6,37 @@ use std::io::BufReader;
...
 
-#[derive(Deserialize, Debug)]
+#[derive(Deserialize, Debug, Default, Clone)]
 pub struct Engines {
     pub node: Option<Requirement>,
     pub yarn: Option<Requirement>,
     pub npm: Option<Requirement>,
 }
 
+#[derive(Deserialize, Debug, Default, Clone)]
```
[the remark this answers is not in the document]
joshwlewis: I actually had that in a previous commit, but I took it out, since other package.json fields tend to favor `camelCase`. Since the casing isn't consistent, I think I'd rather do this case-by-case.

Cargo.toml
```diff
@@ -2,6 +2,7 @@
 members = [
   "buildpacks/nodejs-engine",
   "buildpacks/nodejs-function-invoker",
+  "buildpacks/nodejs-yarn",
```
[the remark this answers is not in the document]
joshwlewis: Yes, the directories were inconsistent with their naming. I'm moving toward adding the `nodejs` prefix to each as I rewrite.

buildpacks/nodejs-yarn/src/yarn.rs
```diff
@@ -0,0 +1,36 @@
...
+            4 => Ok(Yarn::Yarn4),
+            x => Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                format!("Unknown Yarn major version: {x}"),
+            )),
+        }
+    }
+}
+
+impl fmt::Display for Yarn {
```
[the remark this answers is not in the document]
joshwlewis: Yeah, user-facing output feels like target for `Display`. When I wrote this, I was printing this struct for users. Looks like I'm not anymore, so it's not needed.

common/nodejs-utils/src/package_json.rs
```diff
@@ -6,24 +6,37 @@ use std::io::BufReader;
...
-#[derive(Deserialize, Debug)]
+#[derive(Deserialize, Debug, Default, Clone)]
 pub struct PackageJson {
     pub name: String,
     pub version: Option<Version>,
     pub engines: Option<Engines>,
+    pub scripts: Option<Scripts>,
     pub main: Option<String>,
     pub dependencies: Option<HashMap<String, String>>,
     #[serde(rename = "devDependencies")]
```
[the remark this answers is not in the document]
joshwlewis: I decided to go with a case-by-case flag, since the casing isn't consistent across the ecosystem.

common/nodejs-utils/src/vrs.rs
```diff
@@ -1,6 +1,6 @@
 use node_semver::{Range, Version as NSVersion};
```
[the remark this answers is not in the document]
joshwlewis: For context, I ended up on `vrs`, because `clippy::module_name_repetitions` didn't like my initial `version::Version` and I needed something else. 

Honestly, abbreviations seem very common in rust -- half of the standard library uses abbreviated naming: `fs`, `io`, `cmp`, `ptr`, `rc`, `num`, `fmt`, `ops` and I didn't immediately know what all those are, so it feels like there's precedent.

buildpacks/nodejs-yarn/src/yarn.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub(crate) enum Yarn {
+    Yarn1,
+    Yarn2,
+    Yarn3,
+    Yarn4,
+}
+
+impl Yarn {
```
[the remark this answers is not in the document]
joshwlewis: Given we changed the initilaizer to `from_major`, `YarnMajorVersion::from_major(2.0)` looks weird and confusing now.

I had this struct named `YarnLine` at one point.. is that any better?

buildpacks/nodejs-yarn/src/cfg.rs
```diff
@@ -0,0 +1,127 @@
...
+                entry
+                    .map(|e| !e.file_name().to_string_lossy().starts_with('.'))
+                    .unwrap_or(false)
+            })
+        })
+        .unwrap_or(false)
+}
+
+// Fetches the build scripts from a `PackageJson` and returns them in priority order
+pub(crate) fn get_build_scripts(pkg_json: &PackageJson) -> Option<Vec<String>> {
```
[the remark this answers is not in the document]
joshwlewis: The `Option` here was just to make things simpler at the callsite to detect whether to log `No scripts found`. Agree it's not needed. Fixed in c68a702.

buildpacks/nodejs-yarn/tests/integration_test.rs
```diff
@@ -0,0 +1,179 @@
+#![warn(clippy::pedantic)]
+
+use libcnb_test::{assert_contains, assert_not_contains};
+use test_support::Builder::{Heroku20, Heroku22};
+use test_support::{assert_web_response, test_yarn_app};
+
+#[test]
+#[ignore]
```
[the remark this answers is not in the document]
joshwlewis: Added in 347e2ba.

buildpacks/nodejs-yarn/src/yarn.rs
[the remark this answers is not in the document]
Malax: Honestly, in the end this is not a huge deal. We can also leave it as `Yarn` (hence the "nit" prefix). Agreed, `YarnMajorVersion::from_major(2.0)` is weird. :) Feel free to ignore this comment if you want. :)

buildpacks/nodejs-yarn/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+                    | YarnBuildpackError::YarnVersionUnsupported(_) => {
+                        log_error("Yarn version error", err_string);
+                    }
+                }
+            }
+            err => {
+                log_error("Yarn internal buildpack error", err.to_string());
+            }
+        }
+    }
```
Malax, line 197: Have you considered decoupling the error values from the error messages more? I see you're using `thiserror` to derive some basic traits for your error types. This `on_error` implementation uses the `Display` representation for these errors almost exclusively. This means that the actual error messages for our users are distributed across the code-base (`Display` implementation next to the type declarations + prefixes in `on_error`). 

If we would drop `Display` for our errors, we would be forced to put all error messages in one place (maybe even a dedicated module that takes care of mapping errors to user output), isolating this concern from the rest of the buildpack.

I think this approach has several advantages:

- If we ever want to swap out all error messages conditionally we can do that easily since they're all in one module/function. Use cases for this could be different error messages when the buildpack runs on Heroku, locally or for Salesforce Functions. A more unlikely use-case would be translations to a different (human) language.
- It allows us to craft our error messages holistically, both for individual messages but also all of the possible errors in one buildpack. Since we would then see the whole error message at a glance, it's easier to spot formatting, spelling and grammatical issues. When `Display` is used for nested errors, the topmost `Display` might produce strings like "foo bar error: lolwut error: io error: File not found". Explicitly matching against these error trees will yield better error messages.
- Separation of concerns, UI (which these error strings are) would then be strictly separated. We already decouple error logging by returning `Result` instead of having conditionals + `println!` interwoven with the business logic. Moving also the actual strings away from the implementation into one place would bring this concept to conclusion.

If I remember correctly, @edmorley goes with the same style and [the `errors` module for `heroku/maven`](https://github.com/heroku/buildpacks-jvm/blob/1fd4e44152950657e65d94bb007c00f4fc6e09e9/buildpacks/maven/src/errors.rs) uses this approach too. Some of the JVM buildpacks still rely on `Display` for errrors, but eventually I want to get rid of those as well.

This is not a blocker for this PR, we can easily change this later if we want to. Mainly interested about your thoughts on this topic. :)

buildpacks/nodejs-yarn/src/cfg.rs
```diff
@@ -0,0 +1,110 @@
...
+            contents.any(|entry| {
+                entry
+                    .map(|e| !e.file_name().to_string_lossy().starts_with('.'))
+                    .unwrap_or(false)
+            })
+        })
+        .unwrap_or(false)
+}
+
+// Fetches the build scripts from a `PackageJson` and returns them in priority order
```
Malax, line 26: Nit: these look and feel like docstrings, but are regular comments. Is this by accident or on purpose?

Malax (approved):
Great work @joshwlewis! I left some additional nitpicks, but none are blockers! 👍🏻

buildpacks/nodejs-yarn/src/main.rs
joshwlewis: Yes, I tend to agree with your suggestion. The logic for the error messaging is a bit scattered as it is. The error messages aren't as well composed either, with many error's display trait consuming another error's display trait. 

I worry that having to handle every variant, sub-variant, and sub-variant's sub-variant means a huge list of `match`es if we were to refactor. But it's probably worth a shot. I'll tackle this in another PR or maybe try it in the next buildpack.

buildpacks/nodejs-yarn/src/cfg.rs
joshwlewis: Accident. I'll convert them to docstrings.

PR #251, opened 2022-04-12
Add buildpacks:22 to the test matrix

joshwlewis:
We support heroku-22, so we should test against it.

[W-10343872](https://gus.my.salesforce.com/one/one.app#/alohaRedirect/a07EE00000ZVsEZYA1)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -175,6 +175,7 @@ workflows:
               stack-version:
                 - "18"
                 - "20"
+                - "22"
 
       - package-buildpack:
           matrix:


edmorley (approved):

edmorley:
@joshwlewis This PR added Heroku-22 to the test matrix for the shpec tests, however there are also Cutlass + libcnb-test tests that are only tested on `heroku/buildpacks:20` at the moment:
https://github.com/heroku/buildpacks-nodejs/search?q=buildpacks%3A20

What are your thoughts on testing strategy here? :-)

PR #278, opened 2022-06-12
Switch away from deprecated path-based S3 URLs

edmorley:
S3 URLs where the bucket name is part of the URL path are deprecated.

Instead, it's recommended to use the virtual-hosted style references, where the bucket name is part of the domain. 

The latter allows AWS to use DNS to direct requests directly to the appropriate region's S3 endpoint, rather than having to route via the global S3 endpoint (which AWS describe as being a single point of failure/harder to scale etc).

There is a small chance this may also help with some of the S3 reliability issues seen in [heroku/builder](https://github.com/heroku/builder) Circle CI runs - however even if it doesn't, at least we'll no longer be using the deprecated URLs/endpoints.

See:
https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/
https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html
https://docs.aws.amazon.com/general/latest/gr/s3.html#s3_region

GUS-W-11283397.

### DIFF
--- a/buildpacks/nodejs-engine/CHANGELOG.md
+++ b/buildpacks/nodejs-engine/CHANGELOG.md
@@ -4,9 +4,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Switch away from deprecated path-based S3 URLs
+
 ## [0.8.5] 2022/06/08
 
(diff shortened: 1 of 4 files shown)


edmorley:
@joshwlewis @colincasey Any objection to me triggering a new CNB release?

PR #322, opened 2022-08-15
Add tests for the heroku-22 stack

joshwlewis:
This PR adds testing against the heroku-22 stack, via the `heroku/builder:22` builder. Previously, we tested exclusively against `heroku/buildpacks:20`, which is based on the heroku-20 stack.

I was seeing issues with running heroku-22 containers on Ubuntu 20 hosts. Upgrading the hosts to Ubuntu 22 on Circle seemed to resolve that. Others report the same thing: https://askubuntu.com/a/1410408

GUS-W-11300259


.circleci/config.yml
```diff
@@ -80,6 +81,10 @@ jobs:
       - heroku-buildpacks/install-build-dependencies
       - heroku-buildpacks/install-rust
       - heroku-buildpacks/install-libcnb-cargo
+      - run:
+          name: "Trust heroku/builder:22"
+          command: |
+            pack config trusted-builders add heroku/builder:22
```
edmorley, line 87: libcnb-test uses `--trust-builder` (as of https://github.com/heroku/libcnb.rs/pull/409), so this can be removed once the remaining cutlass tests are converted

joshwlewis: I briefly looked into running these with libcnb-test. I didn't pull the trigger mostly because cutlass has a (albeit simple) mechanism to compile meta-buildpacks and their dependent buildpacks. I don't think libcnb has functionality for this yet?

edmorley: Yeah libcnb-test doesn't have that yet.

(To clarify, I wasn't meaning to convert them now, just more of a potential cleanup once the conversion had occurred in the future)

edmorley (approved):

Issue #371, opened 2022-10-04
Cannot run nodejs-engine with empty object


schneems:
Previously in heroku/heroku-buildpacks-nodejs we're able to have a `package.json` with these contents:

```
{}
```

If you attempt to run this buildpack with those contents, you'll receive an error:

```
===> DETECTING
heroku/nodejs-engine 0.8.9
heroku/ruby          2.0.0
heroku/procfile      2.0.0
===> RESTORING
===> BUILDING

[Heroku Node.js Engine Buildpack]

[Checking Node.js version]
ERROR: failed to build: exit status 1


pack stderr:

[Error: Node.js engine package.json error]
Couldn't parse package.json: Could not parse package.json. missing field `name` at line 1 column 2
ERROR: failed to build: executing lifecycle: failed with status code: 51
```

If it's not too much overhead, I would like to support building with an empty object in the `package.json` file. This will aid Ruby users who need a version of `node.js`.

edmorley:
@schneems Creating a `package.json` should not be needed, see:
https://github.com/heroku/buildpacks-nodejs/blob/e1279b5ceae75b686f52512377b71ead6b0819f8/buildpacks/nodejs-engine/src/main.rs#L52-L57

ie: In your build plan require `node` iff you want the node engine buildpack to run. Then add the node engine buildpack as an `optional = true` buildpack in the Ruby meta-buildpack's detection group order.

colincasey:
nice.  i'm glad that's the case @edmorley because it doesn't feel right to require an empty (or even minimal)`package.json` just to use `node` if your buildpack needs it.  

@schneems are you okay with this issue being closed?

edmorley:
> an empty package.json is valid for the npm spec

As found on:
https://docs.npmjs.com/cli/v8/configuring-npm/package-json#name

> If you plan to publish your package, the most important things in your package.json are the name and version fields as they will be required.
> ...
> If you don't plan to publish your package, the name and version fields are optional.

PR #373, opened 2022-10-12
feat: detect and use explicit node runtime dependency in invoker

colincasey:
Changes the Node.js Function invoker to look for the presence of the runtime package (`@heroku/sf-fx-runtime-nodejs`) declared as a dependency in `package.json`

When found, we skip the creation of the layer that installs the runtime.

When not found, we fallback to the previous behavior where a version of the runtime is installed by the CNB and we now emit a warning that notifies a function author that the runtime should be declared from now on.

If the runtime package is declared but in the wrong section (`devDependencies` instead of `dependencies`) then an error will be thrown with a message for the user to add the package to the proper section.

[W-11854639](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000017vsgyYAA/view?ws=%2Flightning%2Fr%2FADM_Epic__c%2Fa3QEE000000MuDF2A0%2Fview)


edmorley (commented):
Cancelling review/converting PR to draft since CI is failing and changes are still being made - so I presume this was not meant to be reviewed yet?

common/nodejs-utils/src/package_json.rs
```diff
@@ -1,16 +1,20 @@
...
+use std::collections::HashMap;
 use std::fs::File;
 use std::io::BufReader;
 use std::path::Path;
 use thiserror::Error;
 
 #[derive(Deserialize, Debug)]
 pub struct PackageJson {
-    pub name: String,
+    pub name: Option<String>,
```
colincasey, line 11: @schneems this should help with #371 
cc: @joshwlewis @edmorley

colincasey:
@edmorley apologies for the noisy commits.  my IDE was not configured to run linting or clippy and it took me awhile to sort that out 😅

test_support/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+
+pub fn get_function_invoker_buildpacks() -> Vec<BuildpackReference> {
+    vec![
+        BuildpackReference::Other(String::from("heroku/nodejs-engine")),
+        BuildpackReference::Other(String::from("heroku/nodejs-npm")),
+        BuildpackReference::Crate,
+    ]
+}
+
+pub fn get_builder_name(builder: Builder) -> &'static str {
```
joshwlewis, line 21: Maybe it might make sense to implement the [`Display` trait](https://doc.rust-lang.org/std/fmt/trait.Display.html) instead here? That'd give us the ability to use `builder.to_string()` or `println!("Builder: {builder}")` with this type.

test_support/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+pub enum Builder {
+    Heroku20,
+    Heroku22,
+}
+
+pub fn get_function_invoker_buildpacks() -> Vec<BuildpackReference> {
+    vec![
+        BuildpackReference::Other(String::from("heroku/nodejs-engine")),
+        BuildpackReference::Other(String::from("heroku/nodejs-npm")),
+        BuildpackReference::Crate,
```
joshwlewis, line 17: This function only seems relevant for the function invoker buildpack. Additionally,`BuildpackReference::Crate` might point to `heroku/nodejs-engine` when called from that buildpack. Maybe this function belongs only in the buildpack tests that needs it?

buildpacks/nodejs-function-invoker/src/main.rs
```diff
@@ -60,27 +65,44 @@ impl Buildpack for NodeJsInvokerBuildpack {
...
+        }
 
         BuildResultBuilder::new()
             .launch(
                 LaunchBuilder::new()
                     .process(
-                        ProcessBuilder::new(process_type!("web"), "sf-fx-runtime-nodejs")
+                        ProcessBuilder::new(process_type!("web"), "npx")
                             .args(vec![
+                                "sf-fx-runtime-nodejs",
```
joshwlewis, line 105: Should this be dynamic instead? It could either be just the package name as read from `buildpack.toml`, or it could be explicitly the package name and version we found in their `package.json`?

joshwlewis: I'm actually surprised this works at all. Don't we need run `npx @heroku/sf-fx-runtime-nodejs` rather than just `npx sf-fx-runtime-nodejs`?

edmorley (commented):
(Some drive-by comments; I'll leave the main review to Josh :-))

test_support/src/lib.rs
colincasey: yes, it is only relevant to the invoker buildpack which is why it is named `get_function_invoker_buildpacks`.  i was envisioning the `test_support` crate as a place to store reusable utility functions for all the buildpacks in the project so that the actual integration test files have minimal clutter.

buildpacks/nodejs-function-invoker/src/main.rs
colincasey: i'd have to split up the package name from `@heroku/sf-fx-runtime-nodejs` into org/name parts which is easy enough to do.  

it's technically not the name of the package required here though, rather the [`bin` entry](https://github.com/forcedotcom/sf-fx-runtime-nodejs/blob/main/package.json#L69) we specify in `@heroku/sf-fx-runtime-nodejs`.  does that change your opinion here at all?

buildpacks/nodejs-function-invoker/src/main.rs
```diff
@@ -60,27 +65,44 @@ impl Buildpack for NodeJsInvokerBuildpack {
...
+                package_version
+            ));
+        } else {
+            log_warning(
+                "Deprecation",
+                format!("Future versions of the Functions Runtime for Node.js ({0}) will not be auto-detected \
+                and must be added as a dependency in package.json.", package_name)
+            );
+            context.handle_layer(
+                layer_name!("implicit_runtime"),
```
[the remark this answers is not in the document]
colincasey: 👍 reverting

buildpacks/nodejs-function-invoker/src/function.rs
```diff
@@ -1,6 +1,9 @@
 use heroku_nodejs_utils::package_json::{PackageJson, PackageJsonError};
 use libcnb::read_toml_file;
 use libherokubuildpack::toml::toml_select_value;
+#[cfg(test)]
+use serde_json::json;
```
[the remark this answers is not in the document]
colincasey: ah, i didn't realize that.  was just following what clippy was telling me to do.  thanks @edmorley!

buildpacks/nodejs-function-invoker/src/function.rs
```diff
@@ -34,6 +37,48 @@ where
...
 }
 
+pub fn get_declared_runtime_package<P>(
+    app_dir: P,
+    package_name: &String,
+) -> Result<Option<(&String, String)>, ExplicitRuntimeDependencyError>
+where
+    P: Into<PathBuf>,
+{
+    let package_json = PackageJson::read(app_dir.into().join("package.json"));
```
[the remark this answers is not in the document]
colincasey: @edmorley it was intentional to ignore any read errors from `package.json` here.  as the explicit runtime dependency is optional (to support backwards-compatibility) i simply wanted this function to act as a signal that either they do have the dependency or not.

but now that i've moved away from that initial choice to handle the error case for when the runtime is declared in the wrong spot i think i should update those other aspects as well.

as for erroring out if the user specifies the runtime in both `dependencies` and `devDependencies` i think that should be allowed.  the function will still run in that case.  @joshwlewis do you have thoughts on this?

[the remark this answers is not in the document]
edmorley: > if the user specifies the runtime in both dependencies and devDependencies

What do npm/yarn do if a dep is in both places? Do they error?

common/nodejs-utils/src/package_json.rs
colincasey: i did add a changelog entry but i can revert those if that's your recommendation.

buildpacks/nodejs-function-invoker/src/function.rs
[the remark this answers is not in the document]
colincasey: one of them wins.  it's unclear which one should but when i just tried it with two runtimes declared at different versions, the `dev` version won which is not ideal.  things also go screwy if the `devDependencies` are pruned.  i'm leaning more towards your assessment now.  will fix this up tomorrow.

buildpacks/nodejs-function-invoker/src/function.rs
```diff
@@ -34,6 +35,46 @@ where
...
+    match package_json {
+        Ok(pjson) => {
+            let package_in_dependencies =
+                find_declared_runtime_package(package_name, &pjson, |pjson| &pjson.dependencies);
+
+            let package_in_dev_dependencies =
+                find_declared_runtime_package(package_name, &pjson, |pjson| {
+                    &pjson.dev_dependencies
+                });
```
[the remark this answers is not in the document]
colincasey: cut me some slack @edmorley, i'm still trying to find my footing in this language 😜

buildpacks/nodejs-function-invoker/src/main.rs
colincasey: nvm, did some more reading about how `npx` determines the executable and using the full package name is preferred vs. relying on the heuristics used [when no `--package` options are provided](https://docs.npmjs.com/cli/v8/commands/npx#description).  will fix this.

buildpacks/nodejs-function-invoker/src/function.rs
```diff
@@ -34,6 +34,30 @@ where
         .and_then(|path| path.exists().then_some(path).ok_or(MainError::MissingFile))
 }
 
+pub fn get_declared_runtime_package<P>(
```
joshwlewis, line 37: I know I've seen package declarations that look like this:`"@hk/sf-fx-runtime": "npm:@heroku/sf-fx-runtime-nodejs@0.12.0". It doesn't look like this code would support that. This is obviously unlikely, but would that be supportable by using something like `npm ls` instead of parsing the package.json ourselves?

joshwlewis: This isn't a blocking comment, btw. This seems very unlikely and a pretty weird use case.

colincasey: @joshwlewis should we support this edge case?  the first time i'd ever seen that style in the wild is with the `@hk/sf-fx-runtime` declaration which makes me think it's fairly uncommon.

but, if you do think we should support this, the value contains the actual package name so we could do some deconstructing of that value instead of using `npm ls`.

joshwlewis: No, I don't really think it's something to support. I was just thinking `npm ls` might be more authoritative in some less common scenarios, since it'd be leveraging npm's package.json, package-lock.json, npm-shrinkwrap.json parsing.

edmorley (commented):
Changes to the parts I commented on previously look great.

Deferring the main approval to Josh :-)

buildpacks/nodejs-function-invoker/src/main.rs
```diff
@@ -60,27 +63,44 @@ impl Buildpack for NodeJsInvokerBuildpack {
...
-                    .package
-                    .clone(),
-            },
-        )?;
+        get_main(app_dir).map_err(NodeJsInvokerBuildpackError::MainFunctionError)?;
+
+        let declared_runtime_package = get_declared_runtime_package(app_dir, package_name)
+            .map_err(NodeJsInvokerBuildpackError::ExplicitRuntimeDependencyFunctionError)?;
+
+        if let Some((package_name, package_version)) = declared_runtime_package {
```
joshwlewis, line 77: `package_name` is being shadowed here. But I think the outer and inner scopes will both have the same contents: `"@heroku/sf-fx-runtime-nodejs"`. I wonder if `get_declared_runtime_package` needs to return the name at all? Maybe it could be `get_runtime_package_version` or similar? We already know the package name, so I think we could remove one memory small memory allocation and a shadowed variable that way.

buildpacks/nodejs-function-invoker/src/main.rs
```diff
@@ -60,27 +63,44 @@ impl Buildpack for NodeJsInvokerBuildpack {
...
+            log_info(format!(
+                "Runtime declared in package.json: {0}@{1}",
```
joshwlewis, line 79: `package_version` isn't guaranteed to be a version number here. It could be a version range, a url, a path, a github repo, etc.. The `@` symbol might look a bit strange in one of those other scenarios.

```suggestion
                "Function runtime declared in package.json: {0}: {1}",
```

buildpacks/nodejs-function-invoker/src/main.rs
```diff
@@ -115,6 +135,12 @@ impl Buildpack for NodeJsInvokerBuildpack {
                     NodeJsInvokerBuildpackError::RuntimeLayerError(_) => {
                         log_error("Node.js Function Invoker runtime layer error", err_string);
                     }
+                    NodeJsInvokerBuildpackError::ExplicitRuntimeDependencyFunctionError(_) => {
+                        log_error(
+                            "Node.js Function Invoker explicit runtime dependency error",
```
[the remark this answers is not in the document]
joshwlewis: Maybe go with `Node.js  function runtime` to match how the JVM version refers to itself?

https://github.com/heroku/buildpacks-jvm/blob/main/buildpacks/jvm-function-invoker/src/main.rs#L72

PR #382, opened 2022-10-27
fix: use the proper command path for local/global runtime

colincasey:
This is a fix against the changes introduced with [#373](https://github.com/heroku/buildpacks-nodejs/pull/373) where the use of `npx` was not detecting the runtime command (`sf-fx-runtime-nodejs`) already installed globally.

The effect of which was the runtime would be downloaded when the Function application initially starts up which is not ideal since it causes slower boot times + may cause non-deterministic behavior.

[W-11854639](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000017vsgyYAA/view?ws=%2Flightning%2Fr%2FADM_Epic__c%2Fa3QEE000000MuDF2A0%2Fview)

### DIFF
--- a/test_support/src/lib.rs
+++ b/test_support/src/lib.rs
@@ -6,6 +6,7 @@ use std::fmt::Formatter;
 use std::time::Duration;
 
 const PORT: u16 = 8080;
+const TIMEOUT: u64 = 10;
 
 pub enum Builder {
     Heroku20,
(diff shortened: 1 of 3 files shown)


test_support/src/lib.rs
```diff
@@ -6,6 +6,7 @@ use std::fmt::Formatter;
 use std::time::Duration;
 
 const PORT: u16 = 8080;
+const TIMEOUT: u64 = 10;
```
edmorley, line 9: I presume this is because previously the test was already borderline not completing within the timeout? In which case bumping seems worth doing.

Though it seems strange that the timeout now needs bumping, when this PR means the implicit case now should be faster, not slower? Unless just bad luck with CI being slower today than normal etc.

edmorley (approved):

Malax (approved):

test_support/src/lib.rs
colincasey: yes, the integration tests would always timeout intermittently for me locally.  just tweaking this because it seems to give more stability when running them locally and matches the timeout [chosen for the builder workflow](https://github.com/heroku/builder/blob/main/.github/workflows/build-test-publish.yml#L75)

PR #418, opened 2022-12-13
Add heroku/nodejs-corepack leveraging libcnb.rs

joshwlewis:
This PR adds a `heroku/nodejs-corepack` buildpack. A full description of the buildpacks functionality is available in the [README](https://github.com/heroku/buildpacks-nodejs/blob/corepack/buildpacks/nodejs-corepack/README.md). For now, the corepack buildpack will only install `yarn` (though corepack can install `pnpm` and `npm`).

There are also changes to the yarn buildpack -- it will no longer install `yarn` if it's already installed (otherwise, `heroku/nodejs-corepack` + `heroku/nodejs-yarn` would result in two `yarn` installations.

The yarn use case is the primary reason for adding this, but the design should allow adding `pnpm` and `npm` support later without too much trouble.

[W-12131343](https://gus.my.salesforce.com/one/one.app#/alohaRedirect/a07EE00001F2GFiYAN)

### DIFF
--- a/buildpacks/nodejs-corepack/src/cmd.rs
+++ b/buildpacks/nodejs-corepack/src/cmd.rs
@@ -1,0 +1,72 @@
+use heroku_nodejs_utils::vrs::Version;
+use libcnb::Env;
+use std::{
+    path::Path,
+    process::{Command, Stdio},
+};
+
+#[derive(thiserror::Error, Debug)]
+pub(crate) enum Error {
+    #[error("Couldn't start corepack command: {0}")]
+    Spawn(std::io::Error),
+    #[error("Couldn't finish corepack  command: {0}")]
+    Wait(std::io::Error),
+    #[error("Corepack command finished with a non-zero exit code: {0}")]
+    Exit(std::process::ExitStatus),
+    #[error("Corepack output couldn't be parsed: {0}")]
+    Parse(String),
+}
+
+/// Execute `corepack --version` to determine corepack version
+pub(crate) fn corepack_version(env: &Env) -> Result<Version, Error> {
+    let output = Command::new("corepack")
+        .arg("--version")
+        .envs(env)
+        .stdout(Stdio::piped())
+        .spawn()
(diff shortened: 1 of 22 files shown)


buildpacks/nodejs-corepack/src/cmd.rs
```diff
@@ -0,0 +1,72 @@
...
+use std::{
+    path::Path,
+    process::{Command, Stdio},
+};
+
+#[derive(thiserror::Error, Debug)]
+pub(crate) enum Error {
+    #[error("Couldn't start corepack command: {0}")]
+    Spawn(std::io::Error),
+    #[error("Couldn't finish corepack  command: {0}")]
```
Malax, line 12:
```suggestion
    #[error("Couldn't finish corepack command: {0}")]
```

buildpacks/nodejs-corepack/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+        } else {
+            DetectResultBuilder::fail().build()
+        }
+    }
+
+    fn build(&self, context: BuildContext<Self>) -> libcnb::Result<BuildResult, Self::Error> {
+        let pkg_mgr = PackageJson::read(context.app_dir.join("package.json"))
+            .map_err(CorepackBuildpackError::PackageJson)?
+            .package_manager
+            .ok_or(CorepackBuildpackError::PackageManager)?;
```
Malax, line 67: Nit and nothing that blocks this PR: I find `PackageManager` as an error value weird - it doesn't include what's the actual issue in the name.

Malax (approved):
Great stuff in here! :)

buildpacks/nodejs-corepack/src/main.rs
joshwlewis: Fixed in [603921e](https://github.com/heroku/buildpacks-nodejs/pull/418/commits/603921ed2d4d8c71522eec354374133aebae2d13).

PR #420, opened 2022-12-30
docs: correct `engines.node` property path

LayZeeDK:


edmorley (approved):

edmorley:
Thank you for the PR! 

I've rebased to pick up the lint fixes on main (CI was failing due to stricter lints in the recently released Rust version).

PR #451, opened 2023-02-02
Fixup nodejs-corepack for release

joshwlewis:
The last release automation for heroku/nodejs-corepack failed silently (https://github.com/heroku/buildpacks-nodejs/actions/runs/3941678312/jobs/6744391908) due to improper configuration, so the version was never incremented. This should fix that issue so that the next release works properly.

Additionally, nodejs-corepack was not included in the last heroku/nodejs release. This should fix that as well.

### DIFF
--- a/buildpacks/nodejs-corepack/buildpack.toml
+++ b/buildpacks/nodejs-corepack/buildpack.toml
@@ -2,7 +2,7 @@ api = "0.8"
 
 [buildpack]
 id = "heroku/nodejs-corepack"
-version = "0.1.0"
+version = "0.1.1"
 name = "Heroku Node.js Corepack"
 homepage = "https://github.com/heroku/buildpacks-nodejs"
 keywords = ["corepack", "node", "node.js", "nodejs", "javascript", "js"]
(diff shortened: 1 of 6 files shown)


buildpacks/nodejs-corepack/buildpack.toml
```diff
@@ -2,7 +2,7 @@ api = "0.8"
 
 [buildpack]
 id = "heroku/nodejs-corepack"
-version = "0.1.0"
+version = "0.1.1"
```
edmorley, line 5: Does https://github.com/heroku/buildpacks-nodejs/blob/main/buildpacks/nodejs-corepack/CHANGELOG.md need a `0.1.1` entry adding manually too?

edmorley (approved):

buildpacks/nodejs-corepack/buildpack.toml
joshwlewis: No, because `0.1.1` hasn't been released yet. It'll get released after I run the "Release Buildpack" action, at which point a PR will be opened to auto-append the version to the changelog.

PR #488, opened 2023-04-03
Create a `heroku/nodejs-pnpm-install` buildpack

joshwlewis:
This PR adds a new buildpack called `heroku/nodejs-pnpm-install`. It will use `pnpm` to install, cache, link, and prune dependencies for node apps.

This PR assumes that `pnpm` will be installed by another buildpack (like `heroku/nodejs-corepack`, which added `pnpm`installation in #489).

[W-12987692](https://gus.my.salesforce.com/one/one.app#/alohaRedirect/a07EE00001Oc89GYAR)


Malax (approved):
LGTM! :)

buildpacks/nodejs-pnpm-install/src/cfg.rs
colincasey, line 1: minor: why is this file named `cfg.rs`?  everything here seems related to caching.

buildpacks/nodejs-pnpm-install/src/main.rs
```diff
@@ -0,0 +1,131 @@
...
+            .map_err(PnpmInstallBuildpackError::PnpmDir)?;
+        cmd::pnpm_set_virtual_dir(&env, &virtual_layer.path)
+            .map_err(PnpmInstallBuildpackError::PnpmDir)?;
+
+        log_header("Installing dependencies");
+        cmd::pnpm_install(&env).map_err(PnpmInstallBuildpackError::PnpmInstall)?;
+
+        let mut metadata = context.store.unwrap_or_default().metadata;
+        let cache_use_count = cfg::read_cache_use_count(&metadata);
+        if cfg::should_prune_cache(cache_use_count) {
```
colincasey, line 75: what's the rationale behind pruning the cache every `x` builds?  why not every build?

colincasey (approved):

buildpacks/nodejs-pnpm-install/src/main.rs
joshwlewis: From https://pnpm.io/cli/store#prune:

> It is best practice to run pnpm store prune occasionally to clean up the store, but not too frequently. Sometimes, unreferenced packages become required again. This could occur when switching branches and installing older dependencies, in which case pnpm would need to re-download all removed packages, briefly slowing down the installation process.

I'm basically trying to honor the use cases mentioned here. I'm not sure this saves anyone a lot of time, but it'll be nice when it does, and the overhead of this conditional isn't too bad (other than CNB's poor treatment of integers requiring us to store the count as floats).

buildpacks/nodejs-pnpm-install/src/cfg.rs
joshwlewis: I thought I would end up sticking more config-releated lookup functions here. I'll rename it.

buildpacks/nodejs-pnpm-install/src/main.rs
joshwlewis: Oh, and the unpruned artifacts only live in the build+cache layer. So we don't have to worry about extra stuff in the prod image.

PR #489, opened 2023-04-06
Add pnpm support to heroku/nodejs-corepack

joshwlewis:
This is the first step towards enabling support for the pnpm package manager. This change will allow `heroku/nodejs-corepack` to install `pnpm`, just like it already does for `yarn`.

This change is a prerequisite for the implementation of `heroku/nodejs-pnpm` buildpack.


dzuelke (approved):

colincasey (approved):
can't say i like this workaround.  i get why you need it though it seems like a major impediment to working with related buildpack changes. maybe https://github.com/heroku/libcnb.rs/pull/563 will help make this possible once it can be integrated into `libcnb-test`

PR #524, opened 2023-05-16
Implement binary mirroring via GitHub Actions

joshwlewis:
This PR moves most of the Node.js and Yarn mirroring logic from the Nodebin service to GitHub actions. The bash scripts are nearly straight copy-pasta. The new `list_unmirrored_versions` binary is used to assist in fanning out each new version to it's own GitHub actions job.

For now, the secrets are configured to mirror to a staging bucket to verify things are working correctly.

Additionally, this had some other changes that are maybe less neccessary, but came along with this work:

- Renamed the `list_versions` binary to `generate_inventory` so it's a bit more intuitive
- `diff_versions` now returns a single line item per addition or removal, so that changelogs look more consistent in the future (no longer will have concatenated lines like `Added node.js 18.2.0, 18.2.1`)
- Some logic from the various rust binaries was centralized into `distribution.rs`, to make it easier to reuse across binaries.
- Updated shellcheck (I was fighting with shell formatting errors in CI that were more about my local config, but I think this makes sense to keep)
- Made some modules private, since I have no expectation for external usage.


common/nodejs-utils/src/distribution.rs
```diff
@@ -0,0 +1,183 @@
...
+        Regex::new(match self {
+            Self::Node => r"node/(?P<channel>\w+)/(?P<arch>[\w-]+)/node-v(?P<version>\d+\.\d+\.\d+)[\w-]+\.tar\.gz",
+            Self::Yarn => r"yarn/(?P<channel>\w+)/yarn-v(?P<version>\d+\.\d+\.\d+(-[\w\.]+)?)\.tar\.gz",
+        }).map_err(|e| anyhow!("Mirrored release regex error: {e}"))
+    }
+}
+
+fn list_upstream_node_versions() -> anyhow::Result<VersionSet> {
+    nodejs_org::list_releases()?
+        .iter()
```
Malax, line 108: Nit, not blocking at all: This could be `into_iter`.

Malax (approved):
Looks good to me. Not super keen on `anyhow` in general, seems ok here tho.

joshwlewis:
> Not super keen on anyhow in general, seems ok here tho.

I didn't think to write it down anywhere, but my thinking was that my `anyhow` usage would be limited to internal commands. Basically, well formatted and constructed errors for buildpack consumers, and I can be a bit more lazy around errors for commands that are only seen by languages folks.

.github/workflows/mirror.yml
```diff
@@ -0,0 +1,95 @@
...
+  schedule:
+    - cron: '00 1 * * 1-5'
+
+jobs:
+  get-unmirrored-node-versions:
+    name: Get unmirrored Node.js versions
+    runs-on: ubuntu-22.04
+    outputs:
+      versions: ${{ steps.get-unmirrored-versions.outputs.versions }}
+    steps:
```
edmorley, line 13: Would https://github.com/Swatinem/rust-cache help improve the job time for this and the other `get-unmirrored-*` job? (The compile currently takes 2 mins)

edmorley (commented):
Awesome! Excited to be one step closer to sunsetting Nodebin :-)

common/nodejs-utils/src/distribution.rs
joshwlewis: Why is `into_iter()` better here? It consumes the vector?

.github/workflows/mirror.yml
joshwlewis: Ah yeah. I didn't even notice this. I guess I copy-pasted this from something older =P.

joshwlewis: Yeah, definite improvement: https://github.com/heroku/buildpacks-nodejs/actions/runs/5070462443/jobs/9105527006 (under 30 seconds now).

common/nodejs-utils/src/distribution.rs
Malax: Yeah, there is no need to create a "view" to the data when the data will be discarded after the function anyways. With `into_iter` we can work directly with the data, not references to it. In some cases (not here) this can reduce the amount of cloning you need to do in iterator chains.

edmorley:
@joshwlewis I spotted one more cleanup item - the IAM user references to nodebin need updating to the new user for prod+staging sections here:
https://github.com/heroku/languages-team/blob/main/guides/creds.md#aws-resources

joshwlewis:
> I spotted one more cleanup item - the IAM user references to nodebin need updating to the new user for the prod+staging sections

Thanks. Fixed in https://github.com/heroku/languages-team/commit/ff5d2c152ec5c3da41b08e46e51eda6dd1695746.

PR #527, opened 2023-05-18
Change all buildpack registries to dockerhub

joshwlewis:
Updates all buildpacks to be published in Docker Hub (instead of our AWS ECR). 

This PR does not change any `package.toml` references to published buildpacks - that will occur as the part of the buildpack release process.

This is a prerequisite of sorts for https://github.com/heroku/buildpacks-nodejs/pull/524. That PR wants to use AWS credentials for S3 binary mirroring, but we're already using AWS credentials for publishing to ECR. This change should free up of AWS credentials, so the other PR doesn't have to deal with scoping credentials.

### DIFF
--- a/meta-buildpacks/nodejs/CHANGELOG.md
+++ b/meta-buildpacks/nodejs/CHANGELOG.md
@@ -4,6 +4,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- - Change release target from ECR to docker.io/heroku/buildpack-nodejs.
+
 ## [0.6.4] 2023/05/09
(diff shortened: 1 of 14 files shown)


meta-buildpacks/nodejs/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- - Change release target from ECR to docker.io/heroku/buildpack-nodejs.
```
edmorley, line 7:
```suggestion
- Change release target from ECR to docker.io/heroku/buildpack-nodejs.
```

edmorley (approved):

PR #533, opened 2023-05-22
Optimize CI integration tests for runtime

joshwlewis:
The integration tests are so slow... something must be done!

This PR changes the libcnb.rs integration test job to a matrix -- one job for each libcnb.rs buildpack, each of which uses a bigger instance.

This improves CI runtime from around 20 minutes to around 8 minutes.

Inspired largely by @Malax's work in https://github.com/heroku/buildpacks-jvm/pull/478.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
         run: cargo test --locked
 
   find-libcnb-buildpacks:
-    name: "Find libcnb buildpacks"
+    name: Find libcnb buildpacks
     runs-on: ubuntu-22.04
     outputs:
       buildpack_dirs: ${{ steps.find-buildpack-dirs.outputs.buildpack_dirs }}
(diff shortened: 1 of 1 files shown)


edmorley (approved):
Nice!

It looks like the function invoker tests are now the long pole out of the `libcnb-test` integration tests:
https://github.com/heroku/buildpacks-nodejs/actions/runs/5048870189/usage

Looking at those tests, I see there are several testing against Heroku-20, and also upgrading from Heroku-20 to Heroku-22:
https://github.com/heroku/buildpacks-nodejs/blob/main/buildpacks/nodejs-function-invoker/tests/integration_test.rs

Now that Functions have been Heroku-22 only for some time, if you wanted you could drop those Heroku-20 functions tests? (Which might help with the overall end to end time, at least when the slow cutlass tests are converted in the future)

joshwlewis:
> Now that Functions have been Heroku-22 only for some time, if you wanted you could drop those Heroku-20 functions tests?

Yeah, I was already thinking about dropping some of those ツ

PR #540, opened 2023-05-25
Fix yarn active version calculation

joshwlewis:
In #537, I adjusted the version requirement for active versions of yarn. But, the semver parser handled the invalid `>= 1.22 >=4.0.0-rc.40` syntax by disregarding the first half, and using only `>=4.0.0-rc.40` half. So the mirroring logic was missing yarn versions less than 4 for a brief time. The problem was a missing `||` operator. I've also added some tests to make sure it's working correctly.

### DIFF
--- a/common/nodejs-utils/src/distribution.rs
+++ b/common/nodejs-utils/src/distribution.rs
@@ -185,18 +185,39 @@ mod tests {
     }
 
     #[test]
-    fn filter_active_yarn() {
+    fn filter_inactive_yarn() {
         let versions = ["1.20.1", "1.22.19", "3.0.0-rc.1", "3.2.3", "4.0.0-rc.44"]
             .into_iter()
             .map(Version::parse)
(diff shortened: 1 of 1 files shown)


common/nodejs-utils/src/distribution.rs
```diff
@@ -92,7 +92,7 @@ impl Distribution {
     fn active_requirement(self) -> anyhow::Result<Requirement> {
         Requirement::parse(match self {
```
colincasey, line 93: would it be safer to create these as a set of version range objects instead of relying on parsing?  or possibly making the parser smart enough to recognize that the input string was invalid and should have thrown an error here?

joshwlewis: Yeah, probably safer. Quite a bit more verbose and harder to read. I'd also have to open up the `Version` API quite a bit.

colincasey (approved):

PR #573, opened 2023-06-26
Use fixed versions in Node.js buildpacks

colincasey:
In preparation for leveraging release automation, all the buildpacks need:
* to be updated to the same version
* to use the URI format of `libcnb:{buildpack_id}` for local buildpacks references

[GUS-W-13658749](https://gus.lightning.force.com/a07EE00001UkjqiYAB)


meta-buildpacks/nodejs/package.toml
```diff
@@ -2,19 +2,19 @@
 uri = "."
 
 [[dependencies]]
-uri = "docker://docker.io/heroku/buildpack-nodejs-engine@sha256:e17edd5fc68cd7326eab47fb1469c44c122673ee11953c40bc2e6e8e2940ddec"
+uri = "libcnb:heroku/nodejs-engine"
```
joshwlewis, line 5: Won't this break `pack build` without the libcnb changes? I feel like this change should be in the other PR?

colincasey: i split them up because this should work with the existing project outside the context release automation.  but something is off.  i see the `build.sh` scripts don't fail the meta-buildpack job even though the build failed

colincasey: @joshwlewis this should be resolved now.  got rid of most of the custom `build.sh` scripts and changed the cutlass integration tests to just copy the directory produced by `cargo libcnb package` into `target`

joshwlewis (approved):

PR #577, opened 2023-06-28
Fix cutlass integration tests

colincasey:
This fixes the cutlass integration tests which no longer need to use a copy of each meta-buildpack with local dependency references as paths to produce a testable buildpack. These can now use the actual meta-buildpacks and rely on `cargo libcnb package` and `libcnb:{buildpack_id}` references to produce the same result.

This also removes the buildpack packaging tests from the CI workflow which is no longer needed.


joshwlewis (approved):
This looks great. Note that you'll need to remove the `package-buildpack` requirements from the `main` branch protection rule if you haven't already.

PR #623, opened 2023-08-15
Added `nodejs-npm-engine` Buildpack

colincasey:
Initial commit of `nodejs-npm-engine` npm buildpack which replaces partial functionality from the existing bash-based `npm` buildpack.

The `nodejs-npm-engine` buildpack adds support for Heroku's existing `engines.npm` entry in `package.json` which allows an application to specify a version of `npm` to be installed. The `nodejs-engine` buildpack has also been modified to provide `npm` as a default system version. This will ensure that there will either be the system or an application-specified version of `npm` available before installing `npm` modules.

PR #625, opened 2023-08-16
Added `nodejs-npm-install` Buildpack

colincasey:
Initial commit of `nodejs-npm-install` npm buildpack which replaces partial functionality from the existing bash-based npm buildpack.

The `nodejs-npm-install` buildpack adds support for Heroku's installing and caching an applications node modules. The `nodejs-engine` buildpack has also been modified to provide npm as a default system version.

## Example Buildpack Output

```
# Heroku npm Install Buildpack

- Installing node modules
  - Using npm version `6.14.18`
  - Creating npm cache
  - Configuring npm cache directory
  - Running `npm ci "--production=false"`

      added 4 packages in 0.866s
            
  - Done (1.245s)
- Running scripts
  - Running `npm run -s heroku-prebuild`

      executed heroku-prebuild
            
  - Done (0.184s)
  - Running `npm run -s build`
     
      executed build
            
  - Done (0.180s)
  - Running `npm run -s heroku-postbuild`

      executed heroku-postbuild
            
  - Done (0.226s)
- Configuring default processes
  - Adding default web process for `npm start`
- Done (finished in 2.310s)
```


buildpacks/nodejs-npm-install/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+
+        let npm_version = cmd::npm_version(&env).map_err(NpmInstallBuildpackError::NpmVersion)?;
+        log_info(format!("npm version: {npm_version}"));
+
+        let npm_cache_layer = context.handle_layer(layer_name!("npm_cache"), NpmCacheLayer {})?;
+
+        cmd::npm_set_cache_config(&env, &npm_cache_layer.path)
+            .map_err(NpmInstallBuildpackError::NpmSetCacheDir)?;
+
+        cmd::npm_set_no_audit(&env).map_err(NpmInstallBuildpackError::NpmSetNoAudit)?;
```
joshwlewis, line 70: Hmm, we haven't needed this in the bash CNB or the classic buildpack. What's the logic for using it here? I don't have a strong opinion here, but it might not be bad for folks to see stuff like `8 vulnerabilities (1 moderate, 5 high, 2 critical)`.

joshwlewis (commented):
Looking good so far. Left some questions - mostly food for thought at this point.

buildpacks/nodejs-npm-install/src/errors.rs
```diff
@@ -0,0 +1,87 @@
...
+    Application(heroku_nodejs_utils::application::Error),
+    BuildScript(cmd::Error),
+    NpmInstall(cmd::Error),
+    NpmSetCacheDir(cmd::Error),
+    NpmSetNoAudit(cmd::Error),
+    NpmVersion(cmd::Error),
+    PackageJson(PackageJsonError),
+}
+
+pub(crate) fn log_user_errors(error: NpmInstallBuildpackError) {
```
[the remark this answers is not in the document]
colincasey: 100% agree. the errors here are part of the polish that still needs to happen but i would also like to incorporate some of the build output improvements from @schneems as part of that.

buildpacks/nodejs-npm-install/src/main.rs
```diff
@@ -0,0 +1,114 @@
...
+            log_warning(warning.header, warning.body);
+        }
+
+        let package_json = PackageJson::read(app_dir.join("package.json"))
+            .map_err(NpmInstallBuildpackError::PackageJson)?;
+
+        let npm_version = cmd::npm_version(&env).map_err(NpmInstallBuildpackError::NpmVersion)?;
+        log_info(format!("npm version: {npm_version}"));
+
+        let npm_cache_layer = context.handle_layer(layer_name!("npm_cache"), NpmCacheLayer {})?;
```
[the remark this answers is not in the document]
colincasey: interesting. i'll check out slice layers.

buildpacks/nodejs-npm-install/src/main.rs
colincasey: oh, i did this early on when i was trying to figure out testing. i can remove 😅

buildpacks/nodejs-npm-install/src/main.rs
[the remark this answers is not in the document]
colincasey: @joshwlewis just following up on this. i don't think slice layers are the right solution since the spec only references these in the export phase, not the build phase.  

also, does `node_modules` need to be cached? according to the [npm ci](https://docs.npmjs.com/cli/v9/commands/npm-ci) docs:
> If a node_modules is already present, it will be automatically removed before npm ci begins its install.

buildpacks/nodejs-npm-install/src/errors.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+        PackageJsonError::ParseError(e) => {
+            print_error_details(logger, e)
+                .announce()
+                .error(&formatdoc! {"
+                    Failed to parse package.json
+
+                    An unexpected error occurred while parsing package.json. 
+                    
+                    Please {TRY_BUILDING_AGAIN}.
```
joshwlewis, line 68: If we failed to parse the `package.json`, building again is unlikely to help. Maybe the suggestion should be to check the formatting of the file?

buildpacks/nodejs-npm-install/src/errors.rs
```diff
@@ -0,0 +1,169 @@
...
+                    Please {TRY_BUILDING_AGAIN}. 
+                "});
+        }
+        npm::VersionError::Parse(e) => {
+            logger.announce().error(&formatdoc! {"
+                    Failed to parse npm version information
+        
+                    An unexpected error occurred while parsing version information from `{e}`. 
+                    
+                    Please {TRY_BUILDING_AGAIN}.
```
joshwlewis, line 107: Again, I'm not sure that building again with no changes will solve anything for a parse error. I'd suggest a message like "Please check your configuration and try again."

buildpacks/nodejs-npm-install/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        let section = logger.section("Installing node modules");
+        log_npm_version(&env, section.as_ref())?;
+        configure_npm_cache_layer(&context, &env, section.as_ref())?;
+        run_npm_install(&env, section.as_ref())?;
+        let logger = section.end_section();
+
+        let section = logger.section("Running scripts");
+        run_build_scripts(&package_json, &env, section.as_ref())?;
+        let logger = section.end_section();
```
joshwlewis, line 72: Why redeclare this variable when you it could have been made mutable on the first declaration?

buildpacks/nodejs-npm-install/src/npm.rs
```diff
@@ -0,0 +1,69 @@
...
+use libcnb::Env;
+use std::path::PathBuf;
+use std::process::Command;
+
+pub(crate) struct SetCacheConfig<'a> {
+    pub(crate) env: &'a Env,
+    pub(crate) cache_dir: PathBuf,
+}
+
+impl<'a> From<SetCacheConfig<'a>> for Command {
```
joshwlewis, line 11: Why not use a simple function like `fn set_cache_config(env, cache_dir) -> Command` rather than the `From` `impl`? This conversion just feels a little indirect and a surprising primary API.

common/nodejs-utils/src/package_manager.rs
```diff
@@ -0,0 +1,36 @@
+use std::fmt::{Display, Formatter};
+use std::path::PathBuf;
+
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum PackageManager {
```
joshwlewis, line 5: Is there any way we could merge this with the `PackageManager` defined https://github.com/heroku/buildpacks-nodejs/blob/a8738e0a0750b76015f9051c732ec61413c410ca/common/nodejs-utils/src/package_json.rs#L48-L51? It'd be nice if we didn't have two.

buildpacks/nodejs-npm-install/src/npm.rs
colincasey: Maybe.  I was borrowing this style from how @Malax and @edmorley implemented commands for tools like [pack](https://github.com/heroku/libcnb.rs/blob/main/libcnb-test/src/pack.rs#L85) and [docker](https://github.com/heroku/libcnb.rs/blob/main/libcnb-test/src/docker.rs#L69) in `libcnb.rs` and, though it also seemed indirect to me at first, I found it useful for separating the command and arguments from how it's invoked.

buildpacks/nodejs-npm-install/src/main.rs
colincasey: No specific reason, I just prefer to avoid `mut` if it's not necessary and a `let` reassignment works just as well here.

buildpacks/nodejs-npm-install/src/errors.rs
colincasey: I'm not sure what other guidance we could give here though. This would only happen if `npm --version` returned output that was not parsable into semver which is a tooling problem, not a user problem.  Should we suggest they try a different version of npm?

common/nodejs-utils/src/package_manager.rs
colincasey: I looked a this a bit closer and i think these are two separate concepts:

- `heroku_nodejs_utils::package_json::PackageManager` - represents the deserialized `packageManager` field found in a `package.json` file.
- `heroku_nodejs_utils::package_manager::PackageManager` - represents a package manager within the Node.js ecosystem.

I don't think it's necessary for these two things to be merged nor for them to have different names since they exist in different namespaces.

buildpacks/nodejs-npm-install/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+- Initial release
```
edmorley, line 10:
```suggestion

### Added

- Initial release
```

buildpacks/nodejs-npm-install/buildpack.toml
```diff
@@ -0,0 +1,28 @@
...
+
+[[stacks]]
+id = "heroku-22"
+
+[[stacks]]
+id = "io.buildpacks.stacks.bionic"
+
+[metadata]
+[metadata.release]
+[metadata.release.docker]
```
edmorley, line 27:
```suggestion
[metadata.release.image]
```

buildpacks/nodejs-npm-install/buildpack.toml
```diff
@@ -0,0 +1,28 @@
...
+
+[[stacks]]
+id = "heroku-20"
+
+[[stacks]]
+id = "heroku-22"
+
+[[stacks]]
+id = "io.buildpacks.stacks.bionic"
+
```
edmorley, line 24: The wildcard stack makes these other stacks redundant (the pack CLI bug with wildcard stacks was fixed long enough ago we no longer need to work around it)

```suggestion
```

buildpacks/nodejs-npm-install/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+                        BuildPlanBuilder::new()
+                            .provides("node_modules")
+                            .requires("npm")
+                            .requires("node_modules")
+                            .requires("node")
+                            .build(),
+                    )
+                    .build()
+            })
+            .unwrap_or_else(|| DetectResultBuilder::fail().build())
```
edmorley, line 51: This silently ignores IO errors reading package.json, and instead "successfully fails" the build. ie: Instead of the build erroring out with an IO error (which would highlight the issue to the user more clearly), the buildpack will fail to detect and the user will not know why.

edmorley (commented):
(I only had a quick skim through; I'll leave the main review to Josh :-))

buildpacks/nodejs-npm-install/src/main.rs
colincasey: yes, this was one of the questions i had asked in our weekly CNB meeting. should we be failing with an error in detect and, if so, will that error be properly surfaced to the user?

edmorley: iirc the answer in the meeting was "yes" and "yes" :-)

joshwlewis (changes requested):

buildpacks/nodejs-npm-install/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+[![CI][CI BADGE]][CI LINK] [![Registry][Registry BADGE]][Registry LINK]
+
+Heroku's official Cloud Native Buildpack for executing `npm install`.
+
+## How it works
+
+The buildpack will pass detection if:
+ 
+- A `package-lock.json` file is found at the root of the application source.
```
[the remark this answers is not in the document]
colincasey: Ah, good catch. I was initially using the lockfile in detect but then I noticed the bash-based npm buildpack allowed for installs without a lockfile.

buildpacks/nodejs-npm-install/tests/integration_test.rs
```diff
@@ -0,0 +1,175 @@
...
+use test_support::{
+    add_build_script, add_package_json_dependency, nodejs_integration_test,
+    nodejs_integration_test_with_config, update_json_file,
+};
+
+#[test]
+#[ignore = "integration test"]
+fn test_npm_install_with_lockfile() {
+    nodejs_integration_test("./fixtures/npm-project", |ctx| {
+        assert_contains!(ctx.pack_stdout, "# Heroku npm Engine Buildpack");
```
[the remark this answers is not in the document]
colincasey: that is a typo and has been fixed.

buildpacks/nodejs-npm-install/src/npm.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub(crate) with_lockfile: bool,
+}
+
+impl<'a> From<Install<'a>> for Command {
+    fn from(value: Install<'a>) -> Self {
+        let mut cmd = Command::new("npm");
+        if value.with_lockfile {
+            cmd.arg("ci");
+        } else {
+            cmd.args(["install", "--no-package-lock"]);
```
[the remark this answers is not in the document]
colincasey: This was only added to maintain parity with the bash-based buildpack. I would be inclined to remove it but can we safely do that?

buildpacks/nodejs-npm-install/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+        warn_later.warn_now();
+        build_result
+    }
+
+    fn on_error(&self, error: libcnb::Error<Self::Error>) {
+        errors::on_error(error)
+    }
+}
+
+fn run_application_checks(
```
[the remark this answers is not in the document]
colincasey: This was only added to maintain parity with the bash-based buildpack. I would be inclined to remove it but can we safely do that?

buildpacks/nodejs-npm-install/src/npm.rs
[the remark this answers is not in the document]
joshwlewis: I don't have data to know, just a strong hunch: most folks wouldn't notice the difference, since they are using a package-lock.json. For anyone not using a `package-lock.json`, a strong message would make sense.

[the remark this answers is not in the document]
colincasey: dropping this for now, will revisit in a follow up work item.

buildpacks/nodejs-npm-install/src/main.rs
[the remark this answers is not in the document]
colincasey: using `package-lock.json` now.  will revisit in a follow up work item.

buildpacks/nodejs-npm-install/src/errors.rs
```diff
@@ -0,0 +1,212 @@
...
+
+            The {buildpack_name} allows customization of the build process by executing the following scripts \
+            if they are defined in {package_json}:
+            - {heroku_prebuild} 
+            - {heroku_build} or {build} 
+            - {heroku_postbuild}
+
+            An unexpected error occurred while executing {build_script}. See the log output above for more information.
+
+            Ensure that this command runs locally without error and retry your build.
```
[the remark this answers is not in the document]
colincasey: changed to:
> Ensure that this command runs locally without error (exit status = 0) and retry your build.

joshwlewis (approved):

edmorley:
@colincasey I happened to notice whilst looking at #686 that there are a number of CI jobs not marked as a required in the branch protection status checks list - these are the new CI jobs for the buildpack added here and in #623.

colincasey:
thanks @edmorley. added the 4 new jobs to the required status checks list.

buildpacks/nodejs-npm-install/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+
+impl Buildpack for NpmInstallBuildpack {
+    type Platform = GenericPlatform;
+    type Metadata = GenericMetadata;
+    type Error = NpmInstallBuildpackError;
+
+    fn detect(&self, context: DetectContext<Self>) -> libcnb::Result<DetectResult, Self::Error> {
+        let package_json_exists = context
+            .app_dir
+            .join(PackageManager::Npm.lockfile())
```
edmorley, line 39: The variable name is `package_json_exists`, however the file being checked instead seems to be the lockfile? (`PackageManager::Npm.lockfile()`). I presume the latter is correct, so the variable name is what needs changing?

PR #626, opened 2023-08-16
Fix for yarn 2.4.3

colincasey:
After [#617](https://github.com/heroku/buildpacks-nodejs/pull/617) was merged our inventory automation ran and the PR to update the `yarn` inventory with `2.4.3` ([#620](https://github.com/heroku/buildpacks-nodejs/pull/620)) had failing tests that showed that the layout for this version did not match our other distributions.

This PR handles this edge case by renaming the `bin/yarn.js` script to `bin/yarn` which matches the binary we expect to find. This is slightly different the `@yarnpkg/cli-dist` layout which uses a portable shell script at `yarn/bin` that calls `node /path/to/yarn.js`.

### DIFF
--- a/buildpacks/nodejs-yarn/CHANGELOG.md
+++ b/buildpacks/nodejs-yarn/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Added Yarn version 2.4.3.
+
 ## [1.1.4] - 2023-08-10
 
 - No changes.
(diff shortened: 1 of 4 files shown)


colincasey:
While this fix does address the issue with using `2.4.3` as a global `yarn` binary, instead of having to add this special case that executes on every build, we've decided to exclude `2.4.3` from our inventory listing. 

This should not impact applications that use `2.4.3` because in Yarn 2+ you are expected to commit the `.yarn/release/yarn-vX.Y.Z.cjs` file which controls the actual version of Yarn used during a build.  The globally installed Yarn simply operates as a wrapper to invoking the commited release version and that should be compatible with all of our other versions of Yarn in our inventory.

PR #628, opened 2023-08-17
Skip `mirror-distribution` job if there are no versions

colincasey:
Prevents the `mirror-distribution` from running when the versions output is an empty JSON array. Otherwise, the strategy evaluator will fail with:

```
Error when evaluating 'strategy' for job 'mirror-distribution'. heroku/buildpacks-nodejs/.github/workflows/_mirror-distribution.yml@d2b2e4b41bd74186941b08ae182261ac87ff5187 (Line: 49, Col: 18): 
Matrix vector 'version' does not contain any values
```


joshwlewis (approved):
Hah - I did notice this was happening, but it was really working as intended. I figured there was little drawback (just a red x on the actions UI). This is better, thanks!

PR #638, opened 2023-08-19
Update to libcnb 0.14.0

edmorley:
Updates libcnb related crates from 0.13.0 to 0.14.0:
https://github.com/heroku/libcnb.rs/releases/tag/v0.14.0

Dependabot didn't open a PR since it appears to have issues with Cargo workspace root dependencies when using the beta dependency grouping feature. (I'll report this upstream shortly)

I've also updated the `nodejs-corepack` integration tests to use the improved `run_shell_command` helper - which is a much better fit for running one off commands. (This helper existed before, but as of libcnb 0.14.0 now offers even more advantages, since it's able to validate exit code.)

GUS-W-13976470.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ node_modules/
 package-lock.json
 .tool-versions
 
+packaged/
 **/target/
 
 .DS_Store
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 3
 
 [[package]]
-name = "addr2line"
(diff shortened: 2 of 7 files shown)


edmorley:
Quite a reduction in transitive dependencies (`+148 −644`) thanks to the Bollard removal :-)

Malax (approved):
Looks good to me, leaving final word about this to @joshwlewis - but probably not a controversial PR :)

joshwlewis (approved):

PR #612, opened 2023-08-09
Configure npm distribution mirror and inventory syncing

colincasey:
Updates the automation workflows to:

- use reusable workflows to reduce duplication of steps
- adds support for `npm` in the list of distributions to mirror to S3
- adds support for `npm` for synchronizing `inventory.toml` updates
- updates the helper binaries used in the automation to support `npm`

[W-13916385](https://gus.lightning.force.com/a07EE00001XwfoGYAR)

PR #647, opened 2023-09-14
Fix downloads for @yarnpkg/cli-dist

colincasey:
The changes to make this download script work for both `npm` and `yarn` distributions [#612](https://github.com/heroku/buildpacks-nodejs/pull/612) has a bug where the requested distribution (`yarn`) differs from its package name if the version is `>=2` (`@yarnpkg/cli-dist`) and that wasn't accounted for in the name of the download file.

The `curl` command that downloads the tarball should have been saving to `yarn-v{x.y.z}.tar.gz` but was instead saving to `@yarnpkg/cli-dist-v{x.y.z}.tar.gz`. This is causing automation failures (e.g.; https://github.com/heroku/buildpacks-nodejs/actions/runs/6179740345/job/16775160336)

This PR keeps the `distribution_name` and `package_name` separate so this mismatch will no longer happen.

It also:
- adds some extra logging which would have made the error more obvious
- DRYS up the `npm_url`, `tarball_url`, and `downloaded_tarball` which previously were being inlined into commands
- adds retries and timeouts to `curl`

### DIFF
--- a/common/bin/download-verify-npm-package
+++ b/common/bin/download-verify-npm-package
@@ -7,9 +7,9 @@
 set -o pipefail
 set -e
 
-package_name=$1
-if [ "npm" != "${package_name}" ] && [ "yarn" != "${package_name}" ]; then
-	echo "Unrecognized distribution - ${package_name}"
+distribution_name=$1
+if [ "npm" != "${distribution_name}" ] && [ "yarn" != "${distribution_name}" ]; then
+	echo "Unrecognized distribution - ${distribution_name}"
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


common/bin/download-verify-npm-package
```diff
@@ -19,28 +19,32 @@ if [ -z "$package_version" ]; then
...
-
 echo "Determining dist url from ${npm_url}" >&2
-url=$(curl -sSf "${npm_url}" | jq -r '.dist.tarball')
 
-echo "Downloading ${package_name} tarball from ${url} ..." >&2
-curl -sSf -o "./${package_name}-v${package_version}.tar.gz" "${url}"
+tarball_url=$(curl --silent --show-error --fail --retry 5 --retry-all-errors --connect-timeout 10 --max-time 60 "${npm_url}" | jq -r '.dist.tarball')
+echo "Downloading ${package_name} tarball from ${tarball_url} ..." >&2
+
+download_file="./${distribution_name}-v${package_version}.tar.gz"
```
Malax, line 35: Nit, optional and non-blocking: `downloaded_file` seems to be a better name here.

Malax (approved):

common/bin/download-verify-npm-package
colincasey: changed it to `downloaded_tarball`

PR #652, opened 2023-09-21
Add otel tracing via file exports to nodejs-corepack

joshwlewis:
This adds an `initTracer` function for Node.js buildpacks to use, which sets up an OpenTelemetry file exporter for tracing data. 

Additionally, this adds some tracing to the`heroku/nodejs-corepack` buildpack.

### DIFF
--- a/buildpacks/nodejs-corepack/src/main.rs
+++ b/buildpacks/nodejs-corepack/src/main.rs
@@ -15,7 +15,6 @@ use libcnb::generic::GenericPlatform;
 use libcnb::layer_env::Scope;
 use libcnb::{buildpack_main, Buildpack, Env};
 use libherokubuildpack::log::log_header;
-use opentelemetry::global;
 use opentelemetry::trace::Tracer;
 
 #[cfg(test)]
--- a/buildpacks/nodejs-corepack/src/main.rs
+++ b/buildpacks/nodejs-corepack/src/main.rs
@@ -15,7 +15,8 @@ use libcnb::generic::GenericPlatform;
 use libcnb::layer_env::Scope;
 use libcnb::{buildpack_main, Buildpack, Env};
 use libherokubuildpack::log::log_header;
-use opentelemetry::trace::Tracer;
+use opentelemetry::trace::{TraceContextExt, Tracer};
+use opentelemetry::KeyValue;
 
 #[cfg(test)]
 use libcnb_test as _;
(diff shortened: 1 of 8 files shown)


buildpacks/nodejs-corepack/src/main.rs
```diff
@@ -37,60 +40,76 @@ impl Buildpack for CorepackBuildpack {
...
-                                .provides(pkg_mgr)
-                                .build(),
-                        )
-                        .build()
-                },
-            )
-        } else {
-            DetectResultBuilder::fail().build()
-        }
+        let tracer = init_tracer(format!("{}", context.buildpack_descriptor.buildpack.id));
```
colincasey, line 43: this is where i think tighter integration with `libcnb.rs` would be helpful. auto-initializing the tracer (or disabling it if opt-in/out can be configured) and pre-populating a root span with the buildpack id (and maybe some other useful default attributes).

buildpacks/nodejs-corepack/src/main.rs
```diff
@@ -37,60 +40,76 @@ impl Buildpack for CorepackBuildpack {
...
-            layer_name!("mgr"),
-            ManagerLayer {
-                package_manager: pkg_mgr,
-            },
-        )?;
-        let mgr_env = mgr_layer.env.apply(Scope::Build, env);
-        cmd::corepack_prepare(&mgr_env).map_err(CorepackBuildpackError::CorepackPrepare)?;
-
-        BuildResultBuilder::new().build()
+        let tracer = init_tracer(format!("{}", context.buildpack_descriptor.buildpack.id));
```
colincasey, line 70: will this clobber the tracer created during `detect`? or does that file get consumed before this happens?

colincasey (commented):
seems fairly straightforward.  

also interesting is the `opentelemetry` crate allows you to [get a reference to the active span from anywhere](https://docs.rs/opentelemetry/latest/opentelemetry/trace/fn.get_active_span.html).

buildpacks/nodejs-corepack/src/main.rs
joshwlewis: `detect` and `build` are different binaries. They shouldn't be in scope at the same time.

buildpacks/nodejs-corepack/src/main.rs
joshwlewis: Yeah, totally agree! I want to trial it here before bringing it into libcnb.rs for now though.

colincasey (approved):

PR #659, opened 2023-09-25
Write telemetry files in append mode

joshwlewis:
The existing telemetry setup was mistakenly replacing all contents on each write, due to the defaults that come with `File::create`. This should correct the file writer so that contents are appended to the file, rather than replacing the file.

This also changes the writer to essentially`/dev/null` when the telemetry file couldn't be created, previously, it would stream telemetry to stdout when there were file issues, which we don't want.

### DIFF
--- a/common/nodejs-utils/src/telemetry.rs
+++ b/common/nodejs-utils/src/telemetry.rs
@@ -20,8 +20,10 @@ pub fn init_tracer(buildpack_name: impl Into<String>) -> BoxedTracer {
     if let Some(parent_dir) = telem_file_path.parent() {
         let _ = create_dir_all(parent_dir);
     }
+
     let exporter = match File::options()
-        .read(false)
+        .read(true)
+        .write(true)
         .create(true)
         .append(true)
         .truncate(false)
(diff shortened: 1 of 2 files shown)


colincasey (approved):

common/nodejs-utils/src/telemetry.rs
```diff
@@ -3,29 +3,44 @@ use std::{
...
         .join(format!("{}.jsonl", bp_name.replace('/', "_")));
+
+    // Ensure the telemetry dir exists
     if let Some(parent_dir) = telem_file_path.parent() {
         let _ = create_dir_all(parent_dir);
     }
-    let exporter = match File::create(telem_file_path) {
+
+    // Create the telemetry file if it doesn't exist
+    let _ = File::options()
```
colincasey, line 24: @joshwlewis to append for github output in our release actions i use:

```
OpenOptions::new()
  .append(true)
  .open(filename)
```

joshwlewis: Yeah, I'm doing that here: https://github.com/heroku/buildpacks-nodejs/pull/659/files#diff-ebe034d4507bc0c116307c3ec6c58a708d461c5f218af356d27bd438542d5b9fR31. I was trying to separate steps due to issues I was seeing.

PR #663, opened 2023-09-29
Prepare integration tests to run under a single builder

colincasey:
Removes the double tests that were added in several integration suites that would explicitly run against the `20` and `22` stacks and updates the CI workflow to control this.

This also leverages `libcnb@0.15.0` to run our integration tests against the composite buildpacks (`heroku/nodejs` and `heroku/nodejs-function`) so the build order/grouping matches what the user would see instead of manually defining individual buildpacks in tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
     name: Find libcnb buildpacks
     runs-on: ubuntu-22.04
     outputs:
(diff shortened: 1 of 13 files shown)


colincasey:
TODO: switch the CI validations from `libcnb.rs integration tests (./buildpacks/nodejs-corepack)` to `Test ({builder}, {buildpack})` after approval

joshwlewis (approved):

PR #674, opened 2023-10-03
Remove cutlass integration tests

colincasey:
All equivalent checks that were done in the cutlass tests have been moved into the Rust integration tests which use the composite buildpacks to execute.


colincasey:
TODO: remove cutlass tests from required validations

test/specs/node/cutlass/basic_spec.rb
```diff
@@ -1,75 +0,0 @@
...
-
-describe "Heroku's Nodejs CNB" do
-  ["heroku/builder:20", "heroku/builder:22"].each do |builder|
-    describe "with #{builder}" do
-      it "handles a downgrade of node engine" do
-        Cutlass::App.new("simple-function", builder: builder).transaction do |app|
-          set_node_version_in_dir(app.tmpdir, version: "^14.0")
-
-          app.pack_build do |pack_result|
-            expect(pack_result.stdout).to include("Installing Node.js 14.")
```
schneems, line 14: We didn't port this test `set_node_version_in_dir` logic to the new tests. Is that intentional?

buildpacks/nodejs-function-invoker/tests/integration_test.rs
```diff
@@ -89,6 +106,59 @@ fn test_function_with_implicit_runtime_dependency_ts() {
...
+        "accessToken": "",
+        "apexClassId": serde_json::Value::Null,
+        "apexClassFQN": serde_json::Value::Null,
+        "functionName": "",
+        "functionInvocationId": serde_json::Value::Null
+    }));
+
+    let mut json_body = None;
+
+    retry(
```
schneems, line 137: A bit of a tangent down the internals of the node buildpack internals. I think that we could refactor this `retry` function to return a result and clean up this code:

```rust
pub fn muhretry<T, E>(
    attempts: u32,
    retry_delay_in_secs: u64,
    retryable_action: impl Fn() -> Result<T, E>,
) -> Result<T, E> {
    let mut result = retryable_action();
    for _ in 1..attempts {
        if result.is_ok() {
            return result;
        }
        std::thread::sleep(Duration::from_secs(retry_delay_in_secs));
        result = retryable_action();
    }
    result
}
```

Then here it could be:

```rust
    muhretry(DEFAULT_RETRIES, DEFAULT_RETRY_DELAY_IN_SECONDS, || {
        ureq::post(&format!("http://{socket_addr}"))
            .set("Content-Type", "application/json")
            .set("Authorization", "")
            .set("ce-id", &id)
            .set("ce-time", "2020-09-03T20:56:28.297915Z")
            .set("ce-type", "")
            .set("ce-source", "")
            .set("ce-specversion", "1.0")
            .set("ce-sfcontext", &sf_context)
            .set("ce-sffncontext", &ssfn_context)
            .send_json(payload.clone())
    })
    .map(|res| res.into_json().expect("expected response to be json"))
    .expect("request to assert function health check response failed.")
```

Without needing the intermediate `mut json_body = None`. As a bonus this `muhretry` example is shorter than the current `retry`. The only difference is that it guarantees it will call the code at least once, but I think that's a positive thing.

Looks like it was just added in #663 so I think it's a good time to update.

schneems (commented):
I checked the individual tests were ported over. All but the node version ones were. Code  comments on the `retry` helper relate more to #663, but I think are still relevant here.

test/specs/node/cutlass/basic_spec.rb
colincasey: i did port that over, yes.  i mistakenly added it to the previous PR (#663) though.  the test is [here](https://github.com/heroku/buildpacks-nodejs/blob/main/buildpacks/nodejs-engine/tests/integration_test.rs#L30-L49)

buildpacks/nodejs-function-invoker/tests/integration_test.rs
colincasey: 👍 i had considered that but it was already working well in its current form in other places so i left it alone and modified the test instead.  will revisit this decision.

schneems: Thanks. It wasn't a blocker, but I like this format better.

schneems (approved):
The current changes look good. One failing test due to ci.yml still referencing cutlass. Other than that :shipit:

test_support/src/lib.rs
```diff
@@ -82,50 +82,36 @@ pub fn retry<T, E>(
     attempts: u32,
     retry_delay_in_secs: u64,
```
schneems, line 83: [Optional] This function could be more general purpose if it took a Duration instead of an integer. Then you wouldn't have to worry about mixing up attempt numbers and sleep time as the caller would have `Duration::from_secs(1)` (or whatever) in it directly.

The constant naming `DEFAULT_RETRY_DELAY_IN_SECONDS` was a bit of a hint that this could maybe be represented as a more specific type.

PR #676, opened 2023-10-04
Move fixtures to their respective buildpacks

colincasey:
This PR moves each integration test fixture found in the `/test/fixtures` directory to a `{libcnb-buildpack}/tests/fixtures` directory. While not necessary, I think this has the following benefits:

- the scope and coverage for each buildpack is clearer
- it discourages sharing fixtures between buildpacks which is better for isolated tests
- test context is easier to track down by having a fixture located nearby


Malax (approved):
Note: `pnpm-lockl.v7.yaml` and `pnpm-lock.v8.yaml` don't have a newline at the end of the file which can cause diffing problems. IDK if this is an issue with pnpm itself or if the fixtures need to be fixed.

PR #692, opened 2023-10-23
Update buildpack metadata in buildpack.toml

edmorley:
Adjusts the buildpack `name`, `description` and `keywords` in the `buildpack.toml` of all buildpacks to match the style discussed in:
https://github.com/heroku/cnb-builder-images/issues/408

These fields are used by the CNB registry and can also be seen in the output of `pack builder inspect`. They are not used by `pack build` or Kodon. In addition to standardising the metadata, one of the objectives here is to ensure users don't accidentally use component buildpacks when they really wanted the composite buildpack.

The `keywords` field has intentionally been omitted from the functions buildpacks since they are EOL and so don't want to be made more discoverable. (The CNB registry doesn't currently ingest/display the keywords currently, but may in the future.)

In addition, I've marked the older Bash based npm buildpack as deprecated, to try and ensure users don't pick that over the newer buildpacks when browsing the CNB registry (we'll need to publish it one more time before it's deleted from this repo).

Lastly, I've cleaned up the `[metadata]` table/subtables to match the concise style used by the Procfile and Python CNBs.

GUS-W-14121598.

### DIFF
--- a/buildpacks/nodejs-npm-install/CHANGELOG.md
+++ b/buildpacks/nodejs-npm-install/CHANGELOG.md
@@ -10,4 +10,3 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Initial release
- 
--- a/buildpacks/nodejs-corepack/CHANGELOG.md
+++ b/buildpacks/nodejs-corepack/CHANGELOG.md
@@ -7,7 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-- Add support for using corepack to install npm ([#685](https://github.com/heroku/buildpacks-nodejs/pull/685))
+### Added
+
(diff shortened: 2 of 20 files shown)


buildpacks/nodejs-npm-install/CHANGELOG.md
```diff
@@ -10,4 +10,3 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Initial release
-
```
colincasey, line 13: did you mean to remove the new line here?

colincasey (approved):
👍 Thanks @edmorley. Only noticed one minor detail with a changelog file missing an ending new line.

buildpacks/nodejs-npm-install/CHANGELOG.md
edmorley: The newline is still there, the removal is of a stray single space character :-)

I opted to include that change in this PR, since it means the changelog contents can be viewed in the PR diff, making it possible to see why a new changelog entry was not added (since the buildpack is new).

PR #698, opened 2023-10-24
Update support for Yarn 4

joshwlewis:
This PR fixes up our Yarn 4 support. 

The primary change in Yarn 4 that was preventing this buildpack from functioning correctly was the change to the default `enableGlobalCache` setting. In Yarn 2.x and 3.x it had a default of `false`, but in 4.x, it's `true`. When set to `true`, yarn 2+ will look for and install dependencies in the cache at `/home/heroku/.yarn/berry/cache`, which isn't cached between builds nor persisted into the final image. In this PR, we set `enableGlobalCache` to `false` for any build using Yarn 2+, forcing Yarn 2+ to use the cache we set up in `/layers/heroku_nodejs-yarn/deps/cache`. Yarn 2.x and 3.x already have a default of `false`, but we're setting it to false anyway since the buildpack won't function correctly if it's set to `true` by a `.yarnrc.yml`. 

I've added fixtures that test zero-installs, nonzero-installs, pnp linker, and the modules linker all on yarn 4 as well.


colincasey:
@joshwlewis why change the yarn 3 test to yarn 4 instead of keeping the yarn 3 one and adding a new yarn 4 one?

joshwlewis:
@colincasey there were two yarn 3 tests already, and we've already got the entire matrix of zero/nonzero pnp/modules cases covered. Was trying not to grow the list of integration tests.

joshwlewis:
@colincasey I've changed my mind, since the yarn 4 work had some surprising breakages. I've added the yarn 3 test back in and also added another yarn 4 test.

colincasey (approved):

PR #705, opened 2023-10-26
Relax requirement for requiring a lockfile with npm

colincasey:
This adds back the previous behavior for the Functions npm process which allows for building an application without a `package-lock.json`.

Related to [W-14371436](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001dByE3YAK/view)


buildpacks/nodejs-npm-install/README.md
```diff
@@ -8,7 +8,7 @@ Heroku's official Cloud Native Buildpack for executing `npm install`.
 
 The buildpack will pass detection if:
  
-- A `package-lock.json` file is found at the root of the application source.
+- A `package.json` file is found at the root of the application source.
```
joshwlewis, line 11: Doesn't that mean this buildpack will be detected and run in `pnpm` or `yarn` scenarios?

colincasey: No, because it's in the [last group](https://github.com/heroku/buildpacks-nodejs/blob/main/meta-buildpacks/nodejs/buildpack.toml#L43-L61) that we check in the composite node buildpack. The first ordered group that matches should be chosen, right? So this will only come into play if the pnpm and Yarn groups fail detection.

buildpacks/nodejs-npm-install/src/main.rs
```diff
@@ -63,13 +66,24 @@ impl Buildpack for NpmInstallBuildpack {
         let app_dir = &context.app_dir;
         let package_json = PackageJson::read(app_dir.join("package.json"))
             .map_err(NpmInstallBuildpackError::PackageJson)?;
+        let has_lockfile = app_dir.join(PackageManager::Npm.lockfile()).exists();
 
         run_application_checks(app_dir, &warn_later)?;
+        if !has_lockfile {
+            log_warning_later(formatdoc! {"
+                This application is missing the npm lockfile ({package_lock_json}). Without a lockfile dependencies may unexpectedly
```
joshwlewis, line 74: I love this warning!

buildpacks/nodejs-npm-install/src/npm.rs
```diff
@@ -43,12 +43,17 @@ impl<'a> From<Version<'a>> for Command {
...
 }
 
 impl<'a> From<Install<'a>> for Command {
     fn from(value: Install<'a>) -> Self {
         let mut cmd = Command::new("npm");
-        cmd.arg("ci");
+        if value.has_lockfile {
+            cmd.arg("ci");
+        } else {
+            cmd.args(["install", "--no-package-lock"]);
```
joshwlewis, line 55: I still don't love this. I don't think we should be supporting apps without lockfiles. We have a solution for supporting functions without lockfiles already in https://github.com/heroku/buildpacks-nodejs/pull/706.

buildpacks/nodejs-npm-install/src/main.rs
joshwlewis: Maybe this warning or similar could go in heroku/nodejs-engine?

joshwlewis (changes requested):

buildpacks/nodejs-npm-install/src/npm.rs
colincasey: 👍 agreed. going forward with #706

colincasey:
Closing in favor of #706

buildpacks/nodejs-npm-install/README.md
joshwlewis: > No, because it's in the [last group](https://github.com/heroku/buildpacks-nodejs/blob/main/meta-buildpacks/nodejs/buildpack.toml#L43-L61) that we check in the composite node buildpack. The first ordered group that matches should be chosen, right? So this will only come into play if the pnpm and Yarn groups fail detection.

Yes, that's right. Folks using the buildpack outside of the meta-buildpack could potentially see the issue, though.

PR #706, opened 2023-10-26
Rollback to the old npm buildpack for Functions

colincasey:
This reverts the Functions composite buildpack to using the old npm buildpack which allows for building an application without a package-lock.json.

Related to [W-14371436](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001dByE3YAK/view)


meta-buildpacks/nodejs-function/package.toml
```diff
@@ -5,10 +5,7 @@ uri = "."
 uri = "libcnb:heroku/nodejs-engine"
 
 [[dependencies]]
-uri = "libcnb:heroku/nodejs-npm-engine"
-
-[[dependencies]]
-uri = "libcnb:heroku/nodejs-npm-install"
+uri = "../../buildpacks/npm"
```
joshwlewis, line 8: This location may not exist in the future. Should we use the remote uri on dockerhub instead?

joshwlewis (changes requested):

meta-buildpacks/nodejs-function/package.toml
colincasey: Well, the implication of this is we would have to keep the bash-based npm buildpack around a bit longer.

joshwlewis (approved):
I like this solution. It doesn't change our stance of requiring a lockfile for our newer CNBs, and also keeps functions users working well.

PR #710, opened 2023-10-30
Update default Node version to 20.x

colincasey:
As of [2023-10-24](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V20.md#2023-10-24-version-2090-iron-lts-richardlau), Node.js 20.x is now the LTS version.  Previously, this buildpack would default to the latest known Node.js release if one was not provided. Now it will default to the latest Active LTS version.

[W-14390764](https://gus.lightning.force.com/lightning/r/a07EE00001dS3sgYAC/view)

### DIFF
--- a/buildpacks/nodejs-engine/CHANGELOG.md
+++ b/buildpacks/nodejs-engine/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Update default node version to 20.x
+
 ## [2.2.0] - 2023-10-26
 
 - No changes.
--- a/buildpacks/nodejs-engine/CHANGELOG.md
+++ b/buildpacks/nodejs-engine/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Update default node version to 20.x
+- Updated default node version to 20.x
 
 ## [2.2.0] - 2023-10-26
 
--- a/buildpacks/nodejs-engine/src/main.rs
+++ b/buildpacks/nodejs-engine/src/main.rs
@@ -30,6 +30,8 @@ mod layers;
 
 const INVENTORY: &str = include_str!("../inventory.toml");
 
+const LTS_VERSION: &str = "20.x";
+
 pub struct NodeJsEngineBuildpack;
 
 impl Buildpack for NodeJsEngineBuildpack {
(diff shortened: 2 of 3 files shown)


buildpacks/nodejs-engine/tests/integration_test.rs
```diff
@@ -10,7 +10,10 @@ use test_support::{
 #[ignore]
 fn simple_indexjs() {
     nodejs_integration_test("./fixtures/node-with-indexjs", |ctx| {
-        assert_contains!(ctx.pack_stdout, "Detected Node.js version range: *");
+        assert_contains!(
+            ctx.pack_stdout,
+            "Detected Node.js version range: >=20.0.0 <21.0.0-0"
```
edmorley, line 15: This log output seems like it would be confusing to users or support. It's not that the `package.json` specified that version, but that it's the default. It seems the default case should be made to have a different log message?

Whilst this issue is pre-existing, now that the default version isn't the wildcard I think it makes it even more confusing and so important to resolve.

buildpacks/nodejs-engine/src/main.rs
```diff
@@ -70,13 +72,16 @@ impl Buildpack for NodeJsEngineBuildpack {
...
+            Requirement::parse(LTS_VERSION).expect("The default Node.js version should be valid");
+
         let version_range = PackageJson::read(context.app_dir.join("package.json"))
             .map_err(NodeJsEngineBuildpackError::PackageJsonError)
             .map(|package_json| {
                 package_json
                     .engines
                     .and_then(|e| e.node)
-                    .unwrap_or_else(Requirement::any)
+                    .unwrap_or(default_version)
```
edmorley, line 84: I presume this change is because before the buildpack installed the latest supported Node.js (which can be a non-LTS release, depending on what stage of the release lifecycle things are at), but moving forwards the buildpack wants to only default to the LTS releases? If so, perhaps worth mentioning this in the PR description? :-)

buildpacks/nodejs-engine/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Update default node version to 20.x
```
edmorley, line 12: (To keep tenses consistent with both prior entries, plus the `### Changed` headings etc)

```suggestion
- Updated default node version to 20.x
```

buildpacks/nodejs-engine/tests/integration_test.rs
colincasey: Agreed. I've updated the code to branch if there is a version given in `package.json` and:
- if present, the `Detected Node.js version range: ...` message will be shown
- if not present, the message will instead be `Node.js version not specified, using 20.x`

buildpacks/nodejs-engine/src/main.rs
colincasey: Exactly.  And now it's defaulting to v21.x which we don't want.

joshwlewis (approved):

PR #713, opened 2023-11-02
Handle missing lockfiles

joshwlewis:
The npm, yarn, and pnpm buildpacks currently fail to detect if there isn't an appropriate lockfile. This creates problems for users who accidentally forget to include a lockfile, since the messaging is currently unhelpful ("No buildpack groups passed detection"). Instead, we should help users understand the issue. This PR changes the behavior of the `heroku/nodejs-npm-install` buildpack so that if it detects dependencies, but finds no lockfile, it throws an error with helpful messaging during the build.

This means `heroku/nodejs-npm-install` passes detect more liberally. It will now pass detection if there are `dependencies` or `devDependencies`, where it previously strictly relied on the presence of `package-lock.json`. We're relying on the ordering in the meta buildpack to match the less common package managers in a more strict manner prior to the npm group, so lockfile-less apps will end up detecting against the npm group (which probably makes sense, given npm is the most ubiquitous package manager).

Also, I noticed that an app that doesn't have any dependencies or lockfiles can't be built, though we've had that functionality in the past. I've added a new final group that should match in scenarios where there aren't dependencies or lockfiles, so a `index.js` and `package.json` without dependencies can build now without needing to bring in a lockfile or run `npm install` during the build.


common/nodejs-utils/src/package_json.rs
```diff
@@ -25,6 +25,18 @@ pub struct PackageJson {
...
 
+impl PackageJson {
+    #[must_use]
+    pub fn has_dependencies(&self) -> bool {
+        for dep_group in [self.dependencies.as_ref(), self.dev_dependencies.as_ref()] {
+            if dep_group.map_or_else(|| false, |deps| !deps.is_empty()) {
+                return true;
+            }
+        }
+        false
```
colincasey, line 36: I had to stop to unpack the `map_or_else` with the return body logic.  May I suggest the following refactor instead?

```suggestion
        let no_deps = !self.dependencies.as_ref().is_some_and(HashMap::is_empty);
        let no_dev_deps = !self
            .dev_dependencies
            .as_ref()
            .is_some_and(HashMap::is_empty);
        !no_deps || !no_dev_deps
```

common/nodejs-utils/src/application.rs
```diff
@@ -30,6 +30,22 @@ pub fn check_for_multiple_lockfiles(app_dir: &Path) -> Result<()> {
...
+    for manager in [
+        PackageManager::Npm,
+        PackageManager::Pnpm,
+        PackageManager::Yarn,
+    ] {
+        if app_dir.join(manager.lockfile()).exists() {
+            return true;
+        }
+    }
+    false
```
colincasey, line 46: Minor - you could simply this with `any`.  E.g.; 

```suggestion
    [
        PackageManager::Npm,
        PackageManager::Pnpm,
        PackageManager::Yarn,
    ]
    .iter()
    .any(|manager| app_dir.join(manager.lockfile()).exists())
```

buildpacks/nodejs-engine/src/main.rs
```diff
@@ -142,6 +148,34 @@ impl Buildpack for NodeJsEngineBuildpack {
...
+                            formatdoc! {
+                                "
+                                    {err_string}
+
+                                    A lockfile from a supported package manager is required to install Node.js
+                                    dependencies. The package.json for this project specifies dependencies, but
+                                    there isn't a lockfile.
+
+                                    To use npm to install dependencies, run `npm install`. This command will
+                                    genereate a {npm_lockfile:?} lockfile.
```
colincasey, line 163:
```suggestion
                                    generate a {npm_lockfile:?} lockfile.
```

colincasey (changes requested):

meta-buildpacks/nodejs/buildpack.toml
```diff
@@ -60,5 +60,11 @@ optional = true
 id = "heroku/nodejs-npm-install"
 version = "2.2.0"
 
+[[order]]
+
+[[order.group]]
+id = "heroku/nodejs-engine"
```
colincasey, line 66: I'm wondering if this catch-all group should be it's own buildpack?  If the logic is in the node-engine, this will force the need for a `package.json` in the functions composite buildpack too, right?

joshwlewis: Yes, you are right. It would have impacted the functions composite buildpack as it was. I've moved the missing lockfile logic into the `heroku/nodejs-npm-install` buildpack instead.

common/nodejs-utils/src/package_json.rs
joshwlewis: Thanks - I ended up deleting this function in the latest commit anyways.

common/nodejs-utils/src/application.rs
joshwlewis: Thanks - I ended up deleting this function in the latest commit anyways. I've merged it with the multi lockfile check you already had.

common/nodejs-utils/src/package_json.rs
joshwlewis: Whoops, I didn't delete this one. But I did like your `.any` suggestion, which I applied in 44e0adf.

meta-buildpacks/nodejs/buildpack.toml
edmorley: It would be great if we could get rid of the Bash based npm buildpack sooner rather than later (we can then clean up some of the non-libcnb.rs buildpack handling in the release automation etc, amongst other things).

To unblock being able to remove the Bash buildpack without enforcing lockfiles on functions customers, what do you think about making the check conditional on the project not being a functions project? Whilst adding such a conditional isn't ideal, it (a) seems better than having to keep the old Bash buildpack around, (b) it could probably be removed sooner than Functions EOL so long as we handle the Functions changelog/comms side and have let the deprecation notice show up in functions build logs for a bit.

If you take that approach, then the lockfile check could still be run in the npm engine buildpack (if desired; I have no preference either way).

joshwlewis: Luckily, @colincasey has already added comms about the deprecation in the bash buildpack. So, we can move functions to `heroku/nodejs-npm-install` as soon as we think it's been long enough warning perioud. If we try to move functions to `heroku/nodejs-npm-install` any sooner, I think that buildpack would need support for lockfile-less installs, which I'm hoping to avoid.

We could retire the source for the bash buildpack earlier if we wanted, by referencing the bash buildpack artifact on dockerhub in the functions meta-buildpack (that would presume it won't need changes). Or we could move it into it's own repo.

common/nodejs-utils/src/application.rs
```diff
@@ -55,20 +42,48 @@ pub fn warn_prebuilt_modules(app_dir: &Path, _warn_later: &WarnGuard<Stdout>) {
...
 impl Display for Error {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
+            Error::MissingLockfile => {
+                writedoc!(
+                    f,
+                    "
+                        Couldn't determine Node.js package manager. Package manager lockfile not found.
+
+                        A lockfile from a supported package manager is required to install Node.js
```
colincasey, line 66: I may be wrong but not ending this sentence with a trailing `\` will end up producing a `\n` here in the message being output which could end up being displayed funny in a terminal. If so, I'd suggest keeping your paragraphs intact by adding the trailing `\` characters.  E.g.;

```rust
"
    A lockfile from a supported package manager is required to install Node.js \
    dependencies. The package.json for this project specifies dependencies, but \
    there isn't a lockfile.
"
```

common/nodejs-utils/src/application.rs
```diff
@@ -55,20 +42,48 @@ pub fn warn_prebuilt_modules(app_dir: &Path, _warn_later: &WarnGuard<Stdout>) {
...
+
+                        Or, to use yarn to install dependencies, run `yarn install`. This command will
+                        generate a {yarn_lockfile:?} lockfile.
+
+                        Or, to use pnpm to install dependencies, run `pnpm install`. This command will
+                        generate a {pnpm_lockfile:?} lockfile.
+
+                        Ensure the resulting lockfile is committed to the repository, then try again.
+                    ",
+                    npm_lockfile = PackageManager::Npm.lockfile(),
```
colincasey, line 81: For areas that you want highlighting (e.g.; ``run `npm install`.``) or the filename `{npm_lockfile:?}` then you should  wrap these with the value formatter from the output library:

```rust
npm_install = fmt::value("npm install"),
npm_lockfile = fmt::value(PackageManager::Npm.lockfile().display()),
```

colincasey (commented):
@joshwlewis this should work better:
- Functions won't be disrupted 
- The `npm-install` CNB will pass detection if there is a package.json with dependencies (with or without a lockfile) and provide a decent error during build if the lockfile is missing
- The fallback group of just the `node-engine` CNB will allow for projects with no dependencies

That all seems reasonable.  I just had some remaining comments regarding the output.

meta-buildpacks/nodejs/buildpack.toml
colincasey: We could also do what we originally did not want to do and add some metadata to the buildpack plan from the `nodejs-function-invoker` CNB that `npm-install` can use to determine if it should:
- allow installing modules with no lockfile
- run the build hooks in the same way as our classic Node buildpack

It's not ideal but it would allow us to at least remove the bash CNB entirely at the cost of having to support a bit of deprecated logic.

common/nodejs-utils/src/application.rs
joshwlewis: Good call. Fixed in c823276.

common/nodejs-utils/src/application.rs
joshwlewis: Fixed in c823276.

common/nodejs-utils/src/application.rs
```diff
@@ -61,26 +61,31 @@ impl Display for Error {
...
 
-                        To use npm to install dependencies, run `npm install`. This command will
-                        generate a {npm_lockfile:?} lockfile.
+                        To use npm to install dependencies, run {npm_install}. \
+                        This command will generate a {npm_lockfile} lockfile.
 
-                        Or, to use yarn to install dependencies, run `yarn install`. This command will
-                        generate a {yarn_lockfile:?} lockfile.
+                        Or, to use yarn to install dependencies, run {yarn_install}. \
+                        This command willgenerate a {yarn_lockfile} lockfile.
```
colincasey, line 76:
```suggestion
                        This command will generate a {yarn_lockfile} lockfile.
```

colincasey (approved):

PR #714, opened 2023-11-02
Remove unused fixtures

joshwlewis:
These fixtures appear to be unused.


edmorley (approved):

colincasey (approved):
Yes, those got moved into the `npm-install` buildpack but I must have forgot to delete these temporary ones that I put in place so we could migrate off of cutlass.

PR #730, opened 2023-11-20
Switch to Cargo.toml based lint configuration

edmorley:
As of the Cargo included in Rust 1.74, lints can now be configured in `Cargo.toml` across whole crates/workspaces:
https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html
https://doc.rust-lang.org/stable/cargo/reference/manifest.html#the-lints-section
https://doc.rust-lang.org/stable/cargo/reference/workspaces.html#the-lints-table

This reduces the boilerplate, and the chance that we forget to enable lints in some targets. The only thing we need to remember is to add the `[lints] workspace = true` to any new crates in the future.

Making this switch exposed a few places where lints weren't enabled and issues had been missed, which have been fixed now.

Since this feature requires Rust 1.74, the MSRV has also been bumped (however, libcnb 0.16.0 already requires Rust 1.74, so in practice this is a no-op).

GUS-W-14523836.

### DIFF
--- a/buildpacks/nodejs-engine/src/main.rs
+++ b/buildpacks/nodejs-engine/src/main.rs
@@ -1,8 +1,3 @@
-#![warn(unused_crate_dependencies)]
-#![warn(clippy::pedantic)]
-#![warn(clippy::cargo)]
-#![allow(clippy::module_name_repetitions)]
-
 use crate::layers::{DistLayer, DistLayerError, WebEnvLayer};
 use heroku_nodejs_utils::inv::Inventory;
 use heroku_nodejs_utils::package_json::{PackageJson, PackageJsonError};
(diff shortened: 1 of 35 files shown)


buildpacks/nodejs-engine/src/main.rs
```diff
@@ -1,8 +1,3 @@
-#![warn(unused_crate_dependencies)]
-#![warn(clippy::pedantic)]
-#![warn(clippy::cargo)]
```
edmorley, line 3: Intentionally removing the `clippy::cargo` lint, since:
- it only really benefits crates that are being published (and this crate isn't being published)
- it contains some lints that are non-deterministic/potentially noisy (eg https://rust-lang.github.io/rust-clippy/master/index.html#/multiple_crate_versions)

Malax (approved):

PR #734, opened 2023-11-29
Update Node functions runtime to 0.14.4

colincasey:
[W-13579117](https://gus.lightning.force.com/a07EE00001TdTe7YAF)


joshwlewis (approved):
Minimum version change makes sense, given documentation already mentions 20.x minimum: https://developer.salesforce.com/docs/platform/functions/guide/set-up.html. And the CLI itself also rejects Node 14.x.

PR #742, opened 2023-12-01
Node.js Runtime Metrics

colincasey:
This is a rewrite of the [heroku-nodejs-plugin](https://github.com/heroku/heroku-nodejs-plugin) using Node Performance APIs instead of a C++ addon to obtain equivalent details about garbage collection and event loop usage.

[W-14591321](https://gus.lightning.force.com/a07EE00001fe15JYAQ), [Related](https://github.com/heroku/heroku-buildpack-nodejs/pull/1178/)

### DIFF
--- a/buildpacks/nodejs-engine/node_runtime_metrics/metrics_collector.cjs
+++ b/buildpacks/nodejs-engine/node_runtime_metrics/metrics_collector.cjs
@@ -1,0 +1,243 @@
+/**
+ * @typedef {{
+ *     "node.gc.collections": number;
+ *     "node.gc.pause.ns": number;
+ *     "node.gc.old.collections": number;
+ *     "node.gc.old.pause.ns": number;
+ *     "node.gc.young.collections": number;
+ *     "node.gc.young.pause.ns": number;
+ * }} MemoryCounters
+ *
+ * @typedef {{
+ *     "node.eventloop.usage.percent": number;
+ *     "node.eventloop.delay.ms.median": number;
+ *     "node.eventloop.delay.ms.p95": number;
+ *     "node.eventloop.delay.ms.p99": number;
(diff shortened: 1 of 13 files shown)


buildpacks/nodejs-engine/node_runtime_metrics/metrics_collector.cjs
```diff
@@ -0,0 +1,243 @@
+/**
```
joshwlewis, line 1: I think this is the same code as https://github.com/heroku/heroku-buildpack-nodejs/blob/main/metrics/metrics_collector.cjs? Any thoughts on how we could avoid having two copies that we'd need to keep in sync?

colincasey: Assuming we want to keep them in sync? 😄 

But, considering that the previous metrics plugin was remained [(mostly) unmodified](https://github.com/heroku/heroku-nodejs-plugin/commits/main/src/monitor.js) during the several years it was in use, I don't imagine this will be changed that often. And the cost of manual copying is low for a single script with zero dependencies. Setting up a dedicated repository and publishing process seems unnecessary. 

Maybe a simple download & copy script could be added if we find we are making lots of changes?

joshwlewis: > I don't imagine this will be changed that often.

I agree.

> Setting up a dedicated repository and publishing process seems unnecessary.

Yeah, I agree there. Maybe one location could be authoritative (the CNB?) and the other (classic) copy could occasionally be synced? We're already doing something similar with the inventory management tooling.

I'm not hung up on it for this PR, though. We can cross that bridge when we need to update this plugin.

PR #764, opened 2024-01-17
Adjust WEB_CONCURRENCY calculations for memory heavy instances

joshwlewis:
This allows the WEB_MEMORY / WEB_CONCURRENCY selection to be more appropriate on memory heavy instances. This mirrors the work in https://github.com/heroku/heroku-buildpack-nodejs/pull/1196.

[GUS](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001g8anxYAA/view)

### DIFF
--- a/buildpacks/nodejs-engine/CHANGELOG.md
+++ b/buildpacks/nodejs-engine/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Adjusted WEB_MEMORY and WEB_CONCURRENCY calculation to be more appropriate on
+  memory heavy instances.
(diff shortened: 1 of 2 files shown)


colincasey (commented):
Minor detail but, since I'm less steeped in these concurrency settings, I would appreciate if many of the values used in this script were named constants (though I see you've moved away from that by removing things like `DEFAULT_WEB_MEMORY_MB` @joshwlewis).

For example: `cmp::min(129_024, max_bytes / 1_048_576)`. What is `129_024` or `1_048_576` supposed to represent again?  I just looked at them recently for #1196 but have already forgotten most of the context 😅

joshwlewis:
@colincasey I've re-added those constants, and added a few more. I think you are correct, naming them will help us both understand the next time we're looking at this.

colincasey (approved):

PR #767, opened 2024-01-23
Update metrics scripts to be compatible with Node v14.10.+

### LINKED ISSUE #754: "node:timers not found" error during build
The `node:timers` package doesn't appear to have been installed properly, causing some builds to fail. This behavior was introduced [here](https://github.com/heroku/buildpacks-nodejs/pull/742/files) in v2.6.1.

Specifically,

```
Error: Cannot find module 'node:timers'
Require stack:
- /layers/heroku_nodejs-engine/node_runtime_metrics/metrics_collector.cjs
- internal/preload
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:966:15)
    at Function.Module._load (internal/modules/cjs/loader.js:842:27)
    at Module.require (internal/modules/cjs/loader.js:1026:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Object.<anonymous> (/layers/heroku_nodejs-engine/node_runtime_metrics/metrics_collector.cjs:24:25)
    at Module._compile (internal/modules/cjs/loader.js:1138:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1158:10)
    at Module.load (internal/modules/cjs/loader.js:986:32)
    at Function.Module._load (internal/modules/cjs/loader.js:879:14)
    at Module.require (internal/modules/cjs/loader.js:1026:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/layers/heroku_nodejs-engine/node_runtime_metrics/metrics_collector.cjs',
    'internal/preload'
  ]
```

Is there a way to disable this by default? Or maybe debug why the package isn't being found?

colincasey:
To support earlier Node.js version several adjustments needed to be made to the metrics script. These include:
- dropping the `node:` protocol from required core modules
- using `http` instead of `fetch`
- using API signatures compatible across all versions:
  - `PerformanceObserver.observe`
  - `PerformanceEntry.kind` vs. `PerformanceEntry.detail.kind`
  - `cluster.isPrimary` vs. `cluster.isMaster`

This PR also includes changes to the CNB to restrict the layer containing the metrics script to only versions of Node `>=14.10`.

[W-14838686](https://gus.lightning.force.com/lightning/r/a07EE00001iPnjsYAC/view)

Fixes #754

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,7 +122,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [ latest, lts/*, lts/-1 ]
+        # check the minimum node version supported by the metrics script and the latest node version
+        # (assumes the versions between have backwards-compatible APIs)
+        version: [14.10.0, latest]
     name: Test Metrics (${{ matrix.version }})
     runs-on: ubuntu-latest
     steps:
@@ -130,4 +132,4 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.version }}
-      - run: node --test buildpacks/nodejs-engine/node_runtime_metrics/test/metrics.spec.cjs
+      - run: npx mocha buildpacks/nodejs-engine/node_runtime_metrics/test/metrics.spec.cjs
--- a/buildpacks/nodejs-engine/CHANGELOG.md
+++ b/buildpacks/nodejs-engine/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Collect Node.js Runtime Metrics for v14.10.0 and up if the application has opted-in. ([#767](https://github.com/heroku/buildpacks-nodejs/pull/767))
+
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -122,12 +122,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: [ latest, lts/*, lts/-1 ]
+        # check the minimum node version supported by the metrics script and the latest node version
+        # (assumes the versions between have backwards-compatible APIs)
+        version: [14.10.0, latest]
```
joshwlewis, line 127: Since `latest` might resolve to the odd number / temporary releases (like 21.x starting next fall), I still think it's a good idea to put `lts/*` in here.

joshwlewis (approved):
Maybe consider adding `lts/*` to the CI matrix, but it's not blocking.

.github/workflows/ci.yml
colincasey: Yes, and that is deliberate since the APIs used here should be stable between the given range of versions so checking against the earliest and latest versions should cover that. If you're feeling like that's not reliable enough then I can add `lts/*` and `lts/-1` back in.

PR #768, opened 2024-01-23
Restrict metrics layer installation to Node >=14.10

colincasey:
Stacked with PR - #767 

The main functionality of this PR is to modify `nodejs-engine` so that the metrics script is only include when the Node.js version is 14.10 or higher + integration tests to verify that behavior.

This PR also includes some refactoring around code related to obtaining the current Node.js version which is used by both `nodejs-engine` and `nodejs-npm-engine`. This is now available in `nodejs-utils` as `node::get_node_version`.

Some common error handling output was also moved into `nodejs-utils`.

[W-14838686](https://gus.lightning.force.com/lightning/r/a07EE00001iPnjsYAC/view)


buildpacks/nodejs-engine/src/main.rs
```diff
@@ -83,27 +90,34 @@ impl Buildpack for NodeJsEngineBuildpack {
...
-        context.handle_layer(
+        let dist_layer = context.handle_layer(
             layer_name!("dist"),
             DistLayer {
                 release: target_release.clone(),
             },
         )?;
 
+        let env = dist_layer.env.apply(Scope::Build, &Env::from_current());
+        let node_version = node::get_node_version(&env)
```
joshwlewis, line 111: Don't we already know the Node.js version we installed (I think it's in the `target_release` object)? It seems like this might be an extra command we don't need.

colincasey:
Closing this PR. The suggestion you noted @joshwlewis is a better fit. The code to restrict the layer installation + the unit tests have been moved into #767.

PR #789, opened 2024-03-04
Update to Buildpack API 0.10, libcnb.rs 0.19.0

joshwlewis:
This updates the libcnb family of dependencies and makes the following changes to support that change:

- [`StackId` has been removed from `libcnb`](https://github.com/heroku/libcnb.rs/pull/773) and the [Cloud Native Buildpacks Buildpack API 0.10 API](https://github.com/buildpacks/spec/blob/buildpack/0.10/buildpack.md?plain=1#L750-L760) in favor of `Target.os` and `Target.arch`. This buildpack now uses `os` + `arch` to determine cache validity instead of `StackId`.
- Bumps the Buildpack API version to 0.10 for libcnb buildpacks
- Includes `[[targets]]` for all libcnb buildpacks
  - Using `linux/amd64`, since these are all compiled only for that platform.
- A bunch of layer method signatures have changed so that the layer is mutable.

Due to previous issues with Buildpack API 0.10 and lifecycle (https://github.com/buildpacks/lifecycle/issues/1308), we should probably wait until `heroku/builder` is updated with lifecycle 0.19 or greater to ship this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,7 +193,7 @@ dependencies = [
  "iana-time-zone",
  "num-traits",
  "serde",
- "windows-targets 0.52.4",
+ "windows-targets",
 ]
 
 [[package]]
@@ -211,8 +211,8 @@ dependencies = [
  "glob",
  "indoc",
  "lazy_static",
- "libcnb",
- "libherokubuildpack",
+ "libcnb 0.17.0",
+ "libherokubuildpack 0.17.0",
  "regex",
  "serde",
  "sha2",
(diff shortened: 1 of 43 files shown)


colincasey (approved):

colincasey (commented):
This should be fine once you bump the `buildpack.toml` to use `api = 0.10`.

colincasey:
@joshwlewis this probably should have a changelog entry if you're going to be bumping the `api` in `buildpack.toml`

colincasey (changes requested):
Will reapprove after the `api` is changed to `0.10`.

colincasey (approved):

PR #795, opened 2024-03-07
Fix flaky integration tests

colincasey:
Several of the metrics tests are temporal-based and, while we can get decent reliability by tweaking the timeouts used, it's not a guarantee that the chosen timeouts will always be effective which can lead to unwanted flakiness during test runs or test runs that wait longer than necessary.

To balance the tradeoff between wait time and test execution time, this PR introduces `wait_for(condition, timeout)` which will continually poll the state of the given `condition` until it passes or the `timeout` is exceeded. This should help to minimize the wait time required to guarantee the expected outcome.

### DIFF
--- a/test_support/src/lib.rs
+++ b/test_support/src/lib.rs
@@ -6,6 +6,7 @@ use libcnb_test::{
     assert_contains, BuildConfig, BuildpackReference, ContainerConfig, ContainerContext,
     TestContext, TestRunner,
 };
+use std::any::Any;
 use std::net::SocketAddr;
 use std::panic;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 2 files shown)


test_support/src/lib.rs
```diff
@@ -119,6 +120,27 @@ pub fn assert_web_response(ctx: &TestContext, expected_response_body: &'static s
...
+    {
+        let result = panic::catch_unwind(|| {
+            condition();
+        });
+        if result.is_ok() {
+            return;
+        }
+        std::thread::sleep(Duration::from_millis(10));
+    }
+    panic!("timeout exceeded");
```
joshwlewis, line 141: Would we ever see the actual failure message while using `wait_for`? What I mean is, if this check continually fails:

```
assert_not_contains!(
  container.logs_now().stderr,
  "Registering metrics instrumentation"
);
```
Not because of a timeout, but because we changed the message, does that error show up? Or is only the `timeout exceeded` message seen?

joshwlewis: Just thinking it might make more sense to do something like this after the timeout:

```
eprintln!("wait_for timeout exceeded");
condition()
```

So we still get the original assertion messages, and can see that it timed out.

colincasey: Good call.  I'll capture the potential error from the condition and propagate it forward so if the desired outcome wasn't reached within the timeout it will print the failure message from `assert_not_contains!(container.logs_now().stderr, "Registering metrics instrumentation");`

PR #814, opened 2024-04-18
Use artifact inventory

runesoerensen:
This PR introduces support for the artifact inventory code shared with the Go buildpack, and updates the GitHub actions for node.js to write a corresponding `inventory.toml` file in the new format (which includes the `sha256` checksum that is used to verify the download, as well as the upstream URL rather than our own S3 mirrored artifact store).

A few notes:

* The shared inventory code is being [pulled directly from the Go buildpack](https://github.com/heroku/buildpacks-nodejs/pull/814/files#diff-1c40ca553da3605cd6bc186c75dc58896ae0e0114a1bb8479939b219747d3117R10). This will be moved to `libherokubuildpack` at some point.
* The inventory file includes arm64 and amd64 linux artifacts and are resolved based on [the current host OS and arch](https://github.com/heroku/buildpacks-nodejs/pull/814/files#diff-8aedf08d8ee6e6284476c90e1f7457c8e0be666ac0cf4822b5f4785f10737214R91-R93).
* `darwin` artifacts and the `staging` channel releases are no longer included in the inventory.
* As the classic nodejs buildpack uses the current implementation for inventory updates, this PR does not update or remove any of the shared code.
* When/if the classic node.js buildpack if migrated to use the new inventory utilities we can remove the node specific code from this repository.

We may also want to consider migrating NPM and Yarn over to the new format at some point (so we can avoid mirroring, and simplify our codebase), but that's probably out of scope for the time being.

### DIFF
--- a/.github/workflows/inventory.yml
+++ b/.github/workflows/inventory.yml
@@ -7,14 +7,54 @@ on:
 
 jobs:
   update-nodejs-inventory:
-    name: Node.js
-    uses: ./.github/workflows/_update-inventory.yml
-    with:
-      name: Node.js
-      distribution: node
-      buildpack_id: heroku/nodejs-engine
-      buildpack_path: buildpacks/nodejs-engine
-    secrets: inherit
+    name: Update Node.js Inventory
+    runs-on: pub-hk-ubuntu-22.04-small
+    steps:
+      - uses: heroku/use-app-token-action@main
+        id: generate-token
+        with:
+          app_id: ${{ vars.LINGUIST_GH_APP_ID }}
+          private_key: ${{ secrets.LINGUIST_GH_PRIVATE_KEY }}
+
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+
+      - id: install-rust-toolchain
+        name: Install Rust Toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+
+      - name: Rebuild Inventory
+        id: rebuild-inventory
+        run: |
+          delimiter="$(openssl rand -hex 8)"
+          {
+            echo "msg<<${delimiter}"
+            cargo run --bin update_node_inventory buildpacks/nodejs-engine/inventory.toml
+            echo "${delimiter}"
+          } >> $GITHUB_OUTPUT
+
+      - name: Update Changelog
+        run: echo "${{ steps.rebuild-inventory.outputs.msg }}" | xargs -r -I '{}' perl -i -p -e 's/\[Unreleased\]\s+/[Unreleased]\n\n- {}/' buildpacks/nodejs-engine/CHANGELOG.md
+
+      - name: Create Pull Request
+        id: pr
+        uses: peter-evans/create-pull-request@v6
+        with:
+          token: ${{ steps.generate-token.outputs.app_token }}
+          title: "Update Node.js Inventory"
+          commit-message: "Update Inventory for heroku/nodejs\n\n${{ steps.rebuild-inventory.outputs.msg }}"
+          committer: ${{ vars.LINGUIST_GH_APP_USERNAME }} <${{ vars.LINGUIST_GH_APP_EMAIL }}>
+          author: ${{ vars.LINGUIST_GH_APP_USERNAME }} <${{ vars.LINGUIST_GH_APP_EMAIL }}>
+          branch: update-nodejs-inventory
+          body: "Automated pull-request to update heroku/nodejs inventory:\n\n${{ steps.rebuild-inventory.outputs.msg }}"
+
+      - name: Configure PR
+        if: steps.pr.outputs.pull-request-operation == 'created'
+        env:
+          GH_TOKEN: ${{ steps.generate-token.outputs.app_token }}
+        run: gh pr merge --squash --auto "${{ steps.pr.outputs.pull-request-number }}"
 
   update-yarn-inventory:
     name: Yarn
--- a/.github/workflows/inventory.yml
+++ b/.github/workflows/inventory.yml
@@ -22,6 +22,9 @@ jobs:
       - name: Update Rust toolchain
         run: rustup update
 
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2.7.3
+
       - name: Rebuild Inventory
         id: rebuild-inventory
         run: |
--- a/.github/workflows/inventory.yml
+++ b/.github/workflows/inventory.yml
@@ -19,11 +19,8 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v4
 
-      - id: install-rust-toolchain
-        name: Install Rust Toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
+      - name: Update Rust toolchain
+        run: rustup update
 
       - name: Rebuild Inventory
         id: rebuild-inventory
--- a/buildpacks/nodejs-engine/src/main.rs
+++ b/buildpacks/nodejs-engine/src/main.rs
@@ -1,9 +1,9 @@
 use crate::layers::{
     DistLayer, DistLayerError, NodeRuntimeMetricsError, NodeRuntimeMetricsLayer, WebEnvLayer,
 };
-use heroku_nodejs_utils::inv::Inventory;
+use heroku_inventory_utils::inv::{resolve, Arch, Inventory, Os};
 use heroku_nodejs_utils::package_json::{PackageJson, PackageJsonError};
-use heroku_nodejs_utils::vrs::Requirement;
+use heroku_nodejs_utils::vrs::{Requirement, Version};
 use libcnb::build::{BuildContext, BuildResult, BuildResultBuilder};
 use libcnb::data::build_plan::BuildPlanBuilder;
 use libcnb::data::launch::{LaunchBuilder, ProcessBuilder};
@@ -17,6 +17,7 @@ use libcnb_test as _;
 use libherokubuildpack::log::{log_error, log_header, log_info};
 #[cfg(test)]
 use serde_json as _;
+use sha2::Sha256;
 #[cfg(test)]
 use test_support as _;
 use thiserror::Error;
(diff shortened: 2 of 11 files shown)


.github/workflows/inventory.yml
```diff
@@ -7,14 +7,54 @@ on:
...
+        run: |
+          delimiter="$(openssl rand -hex 8)"
+          {
+            echo "msg<<${delimiter}"
+            cargo run --bin update_node_inventory buildpacks/nodejs-engine/inventory.toml
+            echo "${delimiter}"
+          } >> $GITHUB_OUTPUT
+
+      - name: Update Changelog
+        run: echo "${{ steps.rebuild-inventory.outputs.msg }}" | xargs -r -I '{}' perl -i -p -e 's/\[Unreleased\]\s+/[Unreleased]\n\n- {}/' buildpacks/nodejs-engine/CHANGELOG.md
```
colincasey, line 39: Since you're in here, maybe now is a good time to leverage `heroku/keep_a_changelog` since this search-and-replace doesn't conform to the Keep a Changelog spec ;)

buildpacks/nodejs-engine/src/main.rs
```diff
@@ -85,30 +88,32 @@ impl Buildpack for NodeJsEngineBuildpack {
             Requirement::parse(LTS_VERSION).expect("The default Node.js version should be valid")
         };
 
-        let target_release =
-            inv.resolve(&version_range)
-                .ok_or(NodeJsEngineBuildpackError::UnknownVersionError(
-                    version_range.to_string(),
-                ))?;
+        let target_artifact = match (consts::OS.parse::<Os>(), consts::ARCH.parse::<Arch>()) {
```
colincasey, line 91: Are `consts::OS` and `consts::ARCH` equivalent to the `context.target.os` and `context.target.arch` provided by [libcnb.rs](https://github.com/heroku/libcnb.rs/blob/main/libcnb/src/target.rs)?

common/nodejs-utils/src/bin/update_node_inventory.rs
```diff
@@ -0,0 +1,168 @@
...
+use std::{
+    collections::{HashMap, HashSet},
+    env, fs, process,
+};
+
+/// Updates the local node.js inventory.toml with versions published on nodejs.org.
+fn main() {
+    let inventory_path = env::args().nth(1).unwrap_or_else(|| {
+        eprintln!("Usage: update_inventory <path/to/inventory.toml>");
+        process::exit(2);
```
colincasey, line 21: Why do the exit codes in this binary jump all over the place? I see `2`, `1`, `4`, `6`, and `7`. Are these supposed to mean something?

common/nodejs-utils/src/bin/update_node_inventory.rs
```diff
@@ -0,0 +1,168 @@
...
+            list.iter()
+                .map(ToString::to_string)
+                .collect::<Vec<_>>()
+                .join(", ")
+        );
+    });
+}
+
+fn list_upstream_artifacts() -> Result<Vec<Artifact<Version, Sha256>>, anyhow::Error> {
+    let target_version = Version::parse("0.8.6").context("Failed to parse version")?;
```
colincasey, line 77: Minor - `target_version` is a bit of a misnomer.  

```suggestion
let earliest_version = Version::parse("0.8.6").context("Failed to parse earliest Node.js version")?;
```

common/nodejs-utils/src/bin/update_node_inventory.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    list_releases()?
+        .into_iter()
+        .filter(|release| release.version >= target_version)
+        .map(|release| get_release_artifacts(&release))
+        .collect::<Result<Vec<_>>>()
+        .map(|nested| nested.into_iter().flatten().collect())
```
colincasey, line 84: This is one of those cases where you'd think a `flat_map` would be helpful but it isn't. Don't be afraid to fallback to loops if might make the code a bit clearer :)

```rust
let mut artifacts = vec![];
for release in list_releases()? {
  if release.version >= target_version {
    for artifact in get_release_artifacts(&release)? {
      artifacts.push(artifact);
    }
  }
}
Ok(artifacts)
```

common/nodejs-utils/src/bin/update_node_inventory.rs
```diff
@@ -0,0 +1,168 @@
...
+        );
+    });
+}
+
+fn list_upstream_artifacts() -> Result<Vec<Artifact<Version, Sha256>>, anyhow::Error> {
+    let target_version = Version::parse("0.8.6").context("Failed to parse version")?;
+
+    list_releases()?
+        .into_iter()
+        .filter(|release| release.version >= target_version)
```
colincasey, line 81: It would be nice if this accounted for versions already in the inventory so that we don't have to make a ton of requests for `SHASUMS256.txt` files.

common/nodejs-utils/src/bin/update_node_inventory.rs
joshwlewis: I think I did this in the original inventory binaries. I was thinking it'd be easier for the devs using it to trace errors back to their source. I don't think it was a particularly good idea, and I don't think it's used anywhere. Feel free to use the exit codes that makes sense.

joshwlewis (approved):
Direction looks good to me. I tend to agree with Colin's comments, so looks good once those are resolved.

buildpacks/nodejs-engine/src/main.rs
runesoerensen: No, the `Target` fields set by libcnb.rs maps to the environment variables provided by the lifecycle (and in turn, read from the run image): https://buildpacks.io/docs/for-buildpack-authors/concepts/targets/

The values will often be equivalent though (but may differ in some scenarios, e.g. when the build and run images have different architectures), but that's as far as I'm aware that won't happen (at least for buildpacks that require OS/arch specific binaries).

common/nodejs-utils/src/bin/update_node_inventory.rs
runesoerensen: Ah that was a mistake -- fixed in https://github.com/heroku/buildpacks-nodejs/pull/814/commits/3f1d6cfc51819af8f428f6bfbedbdd3fddd888be :)

common/nodejs-utils/src/bin/update_node_inventory.rs
runesoerensen: Updated to use exit code 1 consistently (same as other scripts) https://github.com/heroku/buildpacks-nodejs/pull/814/commits/278eaea37b1846d16b19dc16c5d792d7ad729112

common/nodejs-utils/src/bin/update_node_inventory.rs
runesoerensen: That's a good point -- I've simplified the expression as you suggested in https://github.com/heroku/buildpacks-nodejs/pull/814/commits/59cfa15b722cf1e2ef69944fb6b54d3254bc5c0d

.github/workflows/inventory.yml
```diff
@@ -7,14 +7,54 @@ on:
...
+      - uses: heroku/use-app-token-action@main
+        id: generate-token
+        with:
+          app_id: ${{ vars.LINGUIST_GH_APP_ID }}
+          private_key: ${{ secrets.LINGUIST_GH_PRIVATE_KEY }}
+
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+
+      - id: install-rust-toolchain
```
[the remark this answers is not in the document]
runesoerensen: Fixed by removing this step altogether :) https://github.com/heroku/buildpacks-nodejs/pull/814/commits/28a32f50fce7b1b68efc3a39854b5f390f8e0bea

.github/workflows/inventory.yml
```diff
@@ -7,14 +7,54 @@ on:
...
+        name: Install Rust Toolchain
+        uses: actions-rs/toolchain@v1
```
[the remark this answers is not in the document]
runesoerensen: No particular reason (besides from this being mostly copy-pasted from the equivalent script in `buildpacks-go`. I've changed it to use the same approach as other workflows in this repository: https://github.com/heroku/buildpacks-nodejs/pull/814/commits/28a32f50fce7b1b68efc3a39854b5f390f8e0bea

.github/workflows/inventory.yml
```diff
@@ -7,14 +7,54 @@ on:
...
+        with:
+          toolchain: stable
```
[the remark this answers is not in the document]
runesoerensen: There's definitely a benefit to caching here as well -- I've added the caching action here: https://github.com/heroku/buildpacks-nodejs/pull/814/commits/eaa077938f359a89d8f36eb5074a383fc9962853

.github/workflows/inventory.yml
runesoerensen: I'd be delighted to do this :-) But I'd also rather not mess around with the existing GitHub actions as part of this PR (and it seems the approach we take should be consistent across all actions that modify the changelog?) -- I've opened an issue so we can handle this change in a separate PR https://github.com/heroku/buildpacks-nodejs/issues/820. Hope that works?

runesoerensen: Easy to implement though https://github.com/heroku/buildpacks-dotnet/pull/2 :)

common/nodejs-utils/src/bin/update_node_inventory.rs
runesoerensen: I took a stab at this here: https://github.com/heroku/buildpacks-nodejs/compare/4d5f13b03a25900a1844103281f6f5570b538d3c..7e46492f0403d9bf9d0a5bcb033ed69e80675867 -- basically inlined the functions (to avoid having to pass around the list of local/inventory artifacts), and cloning artifacts when there's one in the inventory that matches the os, arch and version for an upstream release.

Not sure it's super elegant, but readability isn't much worse than before :-) And it completes in 1-2 seconds now (vs about a minute with the previous approach)

colincasey: Just gave it a quick run-through. Seems to work well and I don't find it unreadable 😅

runesoerensen:
The PR is ready to go, but I'll wait to hear back from the SA process!

colincasey (approved):

PR #815, opened 2024-04-19
Add linux/arm64 targets

runesoerensen:
This PR simply adds the linux/arm64 targets required to package and use the various buildpacks in this repository.

It depends on https://github.com/heroku/buildpacks-nodejs/pull/814, which introduces arm64 support through updated inventory logic and version resolution that factors in the host OS and arch.

### DIFF
--- a/buildpacks/nodejs-corepack/CHANGELOG.md
+++ b/buildpacks/nodejs-corepack/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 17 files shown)


joshwlewis (approved):

runesoerensen:
I've added changelog entries as well as [targets for `nodejs-npm-install`](https://github.com/heroku/buildpacks-nodejs/pull/815/commits/566cf5fd55da5315edcee09684cba39aa9c7b266) - I'm assuming the missing linux/arm64 target in that `buildpack.toml` was an oversight. Let me know otherwise!

I've removed the "skip changelog" label, and requesting another review as there are changes to several files since this was initially approved by @joshwlewis.

joshwlewis (approved):
Nice work, let's do it!

PR #830, opened 2024-05-03
Fix for npm permission error

### LINKED ISSUE #800: `Error: EACCES: permission denied, mkdir '/layers/heroku_nodejs-npm-install'` due to NPM writing to `/layers` at runtime
image
I reckon these Images are indeed bug existing.

colincasey:
Registers a launch time layer that sets the following environment variables:

- [`npm_config_cache`](https://docs.npmjs.com/cli/v10/using-npm/config#cache) to write to a temporary directory
  - this indirectly configures the [`logs-dir`](https://docs.npmjs.com/cli/v10/using-npm/config#logs-dir) too since it resides within the cache directory
- [`npm_config-update-notifier`](https://docs.npmjs.com/cli/v10/using-npm/config#update-notifier) to disable the update notifier

Closes #800


buildpacks/nodejs-npm-install/src/main.rs
```diff
@@ -108,11 +114,33 @@ fn run_application_checks(
     application::check_for_singular_lockfile(app_dir).map_err(NpmInstallBuildpackError::Application)
 }
 
+fn set_npm_logs_dir(env: &Env) -> Result<(), NpmInstallBuildpackError> {
+    let logs_dir =
```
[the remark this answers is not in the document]
colincasey: @joshwlewis you're right. I also realized this configuration shouldn't be happening at build time too and moved the npm configuration steps into an exec.d program so it'll only apply to the run environment.

buildpacks/nodejs-npm-install/src/bin/npm_runtime_config.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    if let Err(e) = set_temp_cache_dir() {
+        println!("{e}");
+    }
+}
+
+fn disable_update_notifier() -> Result<NamedOutput, CmdError> {
+    Command::new("npm")
+        .args(["config", "set", "update-notifier", "false"])
```
edmorley, line 31: Does npm support configuring these via env vars instead? If so, would having these as build+launch env vars set during the build be simpler? eg:
https://docs.npmjs.com/cli/v10/using-npm/config#environment-variables

buildpacks/nodejs-npm-install/src/bin/npm_runtime_config.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+fn set_temp_cache_dir() -> Result<NamedOutput, CmdError> {
+    Command::new("npm")
+        .args(["config", "set", "cache"])
+        .arg(create_temp_dir_name("npm.XXXXXXX"))
+        .named_output()
+}
+
+fn create_temp_dir_name(template: &str) -> PathBuf {
```
joshwlewis, line 42: Do we need a randomized name here? It sounds weird to recreate a new one every reboot. It seems like a static name would be fine, and that'd probably shrink the size of this binary on the application image, too.

buildpacks/nodejs-npm-install/src/bin/npm_runtime_config.rs
colincasey: @edmorley you're probably right. i'll give that a try.

buildpacks/nodejs-npm-install/src/bin/npm_runtime_config.rs
colincasey: this was driven more by my own expectations around temp dir creation which usually involves a randomly generated file/folder name. i can drop it if you don't think name collisions will ever be a concern.

buildpacks/nodejs-npm-install/src/bin/npm_runtime_config.rs
colincasey: switched to using env vars. it's much simpler now. not sure what i was thinking with that exec.d script 😅

buildpacks/nodejs-npm-install/src/bin/npm_runtime_config.rs
colincasey: switched to using `/tmp/npm_cache` as the name and dropped the randomization

buildpacks/nodejs-npm-install/src/main.rs
```diff
@@ -90,6 +90,8 @@ impl Buildpack for NpmInstallBuildpack {
         let build_result = configure_default_processes(&package_json, section.as_ref());
         let logger = section.end_section();
 
+        context.handle_layer(layer_name!("npm_runtime_config"), NpmRuntimeConfigLayer {})?;
```
edmorley, line 93: Should the config be set up in the engine buildpack instead, given that's where npm is installed/made ready?

edmorley (commented):
The env var approach looks good to me :-)

Leaving the main review to Josh.

buildpacks/nodejs-npm-install/src/main.rs
edmorley: I meant `nodejs-npm-engine`.

Ah right. `ETOOMANYBUILDPACKS` :-(

joshwlewis (approved):

PR #833, opened 2024-05-09
Add `[[metadata.targets]]` to `heroku/nodejs` for arm64 support

joshwlewis:
We haven't added `[[metadata.targets]]` to `heroku/nodejs` yet, so it will be published as single architecture. Adding that here and updating the changelog to match reality.

### DIFF
--- a/meta-buildpacks/nodejs/CHANGELOG.md
+++ b/meta-buildpacks/nodejs/CHANGELOG.md
@@ -7,11 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [3.1.0] - 2024-05-09
-
 ### Added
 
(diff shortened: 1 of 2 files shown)


meta-buildpacks/nodejs/CHANGELOG.md
```diff
@@ -7,12 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [3.1.0] - 2024-05-09
-
 ### Added
 
 - Support for `arm64` and multi-arch images. ([#815](https://github.com/heroku/buildpacks-nodejs/pull/815))
```
joshwlewis, line 12: Moved this to Unreleased, since we didn't actually release `heroku/nodejs` as multi-arch. We can release again with this change.

runesoerensen (approved):

PR #841, opened 2024-05-13
Adding Heroku-24 to CI test matrix

colincasey:
Adds entries to the CI test matrix for `heroku/builder:24` as well as the supported architectures `arm64` and `amd64`. 

Also adds:
* a couple of unit tests to validate that `DistLayerMetadata` (which is responsible for the installed Node.js binary cache) should be invalidated if there are any differences in the `Artifact` information
* an integration test for each package manager (npm, pnpm, Yarn) that verifies native node modules are always recompiled during both a fresh cache install and restored cache install


edmorley (commented):
CI is currently failing during the setup step (due to missing `-y`, which is needed now multiple packages are being installed, since apt-get prompts if more than one package is being installed). 

For Python this is my install step:
https://github.com/heroku/buildpacks-python/blob/de378063c83e1eec6302e79f93f189162b5bbbbf/.github/workflows/ci.yml#L61-L71

You'll also need to adjust the target triple for the compiled buildpack. For Python I use this:
https://github.com/heroku/buildpacks-python/blob/de378063c83e1eec6302e79f93f189162b5bbbbf/tests/mod.rs#L26-L44

edmorley (approved):
Thank you!

colincasey:
@edmorley thanks for pointing out that target triple configuration. I missed that when I was referencing the python buildpack changes. 

I also added some tests to verify assumptions around native node modules installs. Our current cache invalidation logic for modules should handle target architecture changes fine.

joshwlewis (approved):

PR #852, opened 2024-05-30
Use pub-hk-ubuntu-22.04-arm-small

dzuelke:
Smaller runners are more plentiful, and we no longer need to install Docker, either!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,7 @@ jobs:
 
   rust-integration-test:
     name: ${{ matrix.name }} (${{ matrix.builder_tag }}, ${{ matrix.arch }})
-    runs-on: ${{ matrix.arch == 'arm64' && 'pub-hk-ubuntu-22.04-arm-large' || 'ubuntu-latest' }}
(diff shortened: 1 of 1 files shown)


joshwlewis (approved):

.github/workflows/ci.yml
```diff
@@ -90,14 +90,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      # The beta ARM64 runners don't yet ship with the normal installed tools.
-      - name: Install Docker, Rust and missing development libs (ARM64 only)
+      - name: Install Rust and missing development libs (ARM64 only)
```
edmorley, line 93: @dzuelke All of the rest of these install steps aren't needed either. I removed them and my tests pass here:
https://github.com/heroku/buildpacks-python/pull/204

dzuelke: I suspected as much, but wanted to verify first :) What about Rust though, do we not want our own, latest, version?

edmorley: There's already a `Update Rust toolchain` step later on for all arches :-)

dzuelke: #853

PR #874, opened 2024-07-02
Use `keep_a_changelog_file` crate and `create-github-app-token` GHA

colincasey:

PR #875, opened 2024-07-03
Use actions/create-github-app-token instead of heroku/use-app-token

dzuelke:
The use-app-token action is deprecated: https://github.com/heroku/use-app-token-action/pull/17

334930895-d73c5548-5bd5-4a0b-9102-cad77f556f06

GUS-W-16159724


edmorley:
This was just updated in #874

PR #877, opened 2024-07-04
Update to `libcnb` v0.22 and annotate deprecated APIs

colincasey:
Update buildpacks to `libcnb@0.22` and add `#[allow(deprecated)]` annotations to the old layer traits. Removing these deprecations will be addressed in several follow-up PRs.

Closes #871 

## Follow-up PRs
- [ ] #878 
- [ ] #879
- [ ] #880
- [ ] #881
- [ ] #882
- [ ] #883
- [ ] #884

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,7 @@ updates:
       libcnb:
         patterns:
           - "libcnb"
+          - "libcnb-test"
           - "libcnb-*"
           - "libherokubuildpack"
       rust-dependencies:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,7 +14,6 @@ updates:
         patterns:
           - "libcnb"
           - "libcnb-test"
-          - "libcnb-*"
           - "libherokubuildpack"
       rust-dependencies:
         update-types:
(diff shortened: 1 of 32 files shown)


Cargo.lock
```diff
@@ -213,8 +213,8 @@ dependencies = [
  "glob",
  "indoc",
  "lazy_static",
- "libcnb",
- "libherokubuildpack",
+ "libcnb 0.21.0",
```
edmorley, line 216: (for the future) It would be good to remove the dependency on Ruby commons to prevent double-including different versions of libcnb.rs like this

.github/dependabot.yml
```diff
@@ -13,6 +13,7 @@ updates:
       libcnb:
         patterns:
           - "libcnb"
+          - "libcnb-test"
```
edmorley, line 16: (optional) Yeah I don't know what's up with Dependabot not grouping the dependencies properly when using the wildcard. Good idea to try listing the package names explicitly. Given all of the package names are now listed, the wildcard could be removed too?

edmorley (approved):

Cargo.lock
colincasey: Noted. I'd really like to tackle the inconsistent output across all these CNBs soon but, right now, this doesn't seem to be causing a problem.

PR #892, opened 2024-07-11
Add CI step to docker pull the run image ahead of tests

colincasey:
During the integration tests, if the application is started this will cause `pack` to download the run image if it hasn't already been pulled. Depending on how the tests are executed this could cause multiple attempts to download the run image. 

To be more respectful towards Docker pull limits, this PR attempts to pull the run image using metadata from the builder. This happens before the tests execute and only for tests that attempt to start the container so that pulls should be minimized.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [ "main" ]
+    branches: ["main"]
   pull_request:
 
 permissions:
@@ -102,6 +102,8 @@ jobs:
         uses: buildpacks/github-actions/setup-pack@v5.7.2
       - name: Pull builder image
         run: docker pull ${{ env.INTEGRATION_TEST_CNB_BUILDER }}
+      - name: Pull run image
+        run: docker pull $(pack builder inspect ${{ env.INTEGRATION_TEST_CNB_BUILDER }} -o json | jq -r '.remote_info.run_images[0].name')
       - name: Run integration tests
         working-directory: ${{ matrix.dir }}
         run: cargo test --locked -- --ignored --test-threads 16
@@ -134,7 +136,7 @@ jobs:
       matrix:
         # check the minimum node version supported by the metrics script and the latest node version
         # (assumes the versions between have backwards-compatible APIs)
-        version: [ 14.10.0, latest ]
+        version: [14.10.0, latest]
     name: Test Metrics (${{ matrix.version }})
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -102,6 +102,12 @@ jobs:
         uses: buildpacks/github-actions/setup-pack@v5.7.2
       - name: Pull builder image
         run: docker pull ${{ env.INTEGRATION_TEST_CNB_BUILDER }}
+      - name: Check for tests that start the container
```
edmorley, line 105: The tests that start the container don't actually do any image pulling. The app image will have already been built at that point, by the `pack build` run by any integration test (whether it uses `start_container` or not).

ie: The run image should always be pulled for any integration test, rather than only some of them.

.github/workflows/ci.yml
```diff
@@ -102,6 +102,12 @@ jobs:
...
+        id: needs-run-image
+        run: echo "result=$(grep --quiet --recursive "start_container" ${{ matrix.dir }}/tests && echo "true" || echo "false")" >> $GITHUB_OUTPUT
+      - name: Pull run image
+        if: steps.needs-run-image.outputs.result == 'true'
+        run: docker pull $(pack builder inspect ${{ env.INTEGRATION_TEST_CNB_BUILDER }} -o json | jq -r '.remote_info.run_images[0].name')
```
edmorley, line 110: (optional) The `pack builder inspect` command currently generates metadata for both the local builder image (if it exists) and also the remote builder image on the registry. This is a slightly puzzling implementation choice on Pack CLI's side. There doesn't appear to be a way to disable the remote image querying.

As such, using `pack builder inspect` results in additional Docker Hub calls compared to using `docker inspect`. Since these will only be once per CI job in this step (rather than once per `libcnb-test` integration test) then the impact could be worse - however, for Python I decided to use `docker inspect` instead:
https://github.com/heroku/buildpacks-python/pull/223

I don't mind which you use in this repo though - up to you :-)

colincasey: Yeah, I saw your PR and will adopt the same technique. As always, you find the optimal way @edmorley 😄

.github/workflows/ci.yml
colincasey: Ah, okay. This I wasn't sure about. Thanks for the clarification.

colincasey: Dropped this check

.github/workflows/ci.yml
colincasey: Fixed to use `docker inspect`

PR #893, opened 2024-07-15
Switch to the new GitHub Actions runner groups

edmorley:
To update to Ubuntu 24.04, and to prevent the rate limit issues mentioned here:
https://github.com/heroku/buildpacks-go/pull/283

See also:
https://salesforce.quip.com/bu6UA0KImOxJ#temp:C:GZR57becb9df8d94f80b132168fd
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

Closes #888.
GUS-W-16238120.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ permissions:
 jobs:
 
   shell-lint:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     container:
       image: koalaman/shellcheck-alpine:v0.9.0
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -79,7 +79,7 @@ jobs:
 
   rust-integration-test:
     name: ${{ matrix.name }} (${{ matrix.builder_tag }}, ${{ matrix.arch }})
-    runs-on: ${{ matrix.arch == 'arm64' && 'pub-hk-ubuntu-22.04-arm-small' || 'ubuntu-latest' }}
+    runs-on: ${{ matrix.arch == 'arm64' && 'pub-hk-ubuntu-24.04-arm-medium' || 'ubuntu-24.04' }}
```
edmorley, line 82: I've switched to medium since:
1. The cargo test invocation uses `--test-threads 16` which would benefit from more CPUs
2. It helps to offset the increased boot time of the custom runners (which don't benefit from the same large Slack pool as the standard runners)

colincasey (approved):
thanks for the PR @edmorley

PR #926, opened 2024-09-24
Migrate to libherokubuildpack inventory

runesoerensen:
This PR removes the `heroku-inventory-utils` dependency in favor of the [recently migrated](https://github.com/heroku/libcnb.rs/pull/861) `inventory` code in `libherokubuildpack`.

### DIFF
--- a/buildpacks/nodejs-engine/src/install_node.rs
+++ b/buildpacks/nodejs-engine/src/install_node.rs
@@ -2,7 +2,6 @@ use std::fs;
 use std::io::Read;
 use std::path::Path;
 
-use heroku_inventory_utils::inv::Artifact;
 use libcnb::build::BuildContext;
 use libcnb::data::layer_name;
 use libcnb::layer::{
@@ -10,6 +9,7 @@ use libcnb::layer::{
 };
 use libherokubuildpack::download::download_file;
 use libherokubuildpack::fs::move_directory_contents;
+use libherokubuildpack::inventory::artifact::Artifact;
 use libherokubuildpack::log::log_info;
 use libherokubuildpack::tar::decompress_tarball;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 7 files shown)


buildpacks/nodejs-engine/src/install_node.rs
```diff
@@ -23,7 +23,7 @@ use crate::{NodeJsEngineBuildpack, NodeJsEngineBuildpackError};
 
 pub(crate) fn install_node(
     context: &BuildContext<NodeJsEngineBuildpack>,
-    distribution_artifact: &Artifact<Version, Sha256>,
+    distribution_artifact: &Artifact<Version, Sha256, Option<()>>,
```
colincasey, line 26: `Option<()>` is for the metadata type on the artifact, right? Is it possible to use `()` directly since we don't (yet) support any metadata on the Node.js artifacts?

common/nodejs-utils/src/bin/update_node_inventory.rs
```diff
@@ -47,11 +41,11 @@ fn main() -> Result<()> {
 
 fn write_inventory(
     inventory_path: impl Into<PathBuf>,
-    upstream_artifacts: &HashSet<Artifact<Version, Sha256>>,
+    upstream_artifacts: &[Artifact<Version, Sha256, Option<()>>],
```
colincasey, line 44: Why doesn't `Artifact` implement `Hash`?

buildpacks/nodejs-engine/src/install_node.rs
runesoerensen: I had a similar question :-) See this discussion/PR for more details https://github.com/Malax/inventory/pull/3

common/nodejs-utils/src/bin/update_node_inventory.rs
runesoerensen: It was [removed a while back](https://github.com/Malax/inventory/commit/502319a1a166216d9d71b9bb58ccc8749a17f5d4). I implemented in the go inventory, but it was admittedly a bit of a hack, and not particularly intuitive that an `Artifact` would automatically implement `Hash` especially based on the checksum value alone.

colincasey (approved):

Issue #938, opened 2024-10-17
Migrate classic buildpack to artifact inventory code


mlarraz:
The changes in https://github.com/heroku/buildpacks-nodejs/pull/814 intentionally skipped these files used exclusively by the classic buildpack:

* https://github.com/heroku/buildpacks-nodejs/blob/main/common/nodejs-utils/src/bin/generate_inventory.rs
* https://github.com/heroku/buildpacks-nodejs/blob/main/common/nodejs-utils/src/bin/resolve_version.rs

Migrating these is a blocker to allowing the classic buildpack to support ARM64: https://github.com/heroku/heroku-buildpack-nodejs/issues/1268

The `generate_inventory` script appears to have been superseded by the `update_node_inventory`, but this does not work on the classic buildpack because it expects the Changelog entries to follow strict semver, which [it does not](https://github.com/heroku/heroku-buildpack-nodejs/blob/main/CHANGELOG.md#v266---2024-10-04).

PR #940, opened 2024-10-17
Update resolve_version to support both inventory formats

mlarraz:
First step at addressing https://github.com/heroku/buildpacks-nodejs/issues/938

This preserves the existing script's behavior while also allowing it to read the newer Inventory format from `heroku_inventory_utils` as used in https://github.com/heroku/buildpacks-nodejs/blob/main/buildpacks/nodejs-engine/src/main.rs.

I also added a more detailed error message for when no resolution is found, listing the requirements as well as target OS and arch.


mlarraz:
Updated to incorporate changes from https://github.com/heroku/buildpacks-nodejs/pull/926

colincasey:
Going to mark this as a draft for now due to the reasons mentioned in https://github.com/heroku/heroku-buildpack-nodejs/pull/1321#issuecomment-2688044634. Will need to revisit at a later date but, for now, there is no bandwidth for these changes.

PR #997, opened 2025-01-09
Fix warning for checked in `node_modules` dir

colincasey:
One of the features of the `commons` logger that was removed in https://github.com/heroku/buildpacks-nodejs/pull/993 was the ability to collect log messages that would be emitted at a later point. This feature is not available in `bullet_stream` so when I moved the logging message that notifies the user they've checked in their `node_modules` directory to the end of the build phase, I mistakenly moved the timing of that check to run at the end of the build phase as well.

This check needs to run at the beginning of the build phase and the warning needs to be emitted at the end. This PR contains changes to do this + adds an integration test to verify this case.


colincasey:
Closing this in favor of #998

PR #998, opened 2025-01-09
Remove check and warning for `node_modules` directory

colincasey:
This warning was carried over from the classic Heroku Node.js buildpack which did support both `npm install` and `npm ci`. If a `node_modules` directory existed and there was no lockfile present, this could cause issues with the dependencies installed.

This buildpack only supports `npm ci` so this check and warning is not required. If the `node_modules` directory is already present, `npm ci` will automatically remove it.

### DIFF
--- a/buildpacks/nodejs-npm-install/CHANGELOG.md
+++ b/buildpacks/nodejs-npm-install/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+
+- Check and warning for `node_modules` directory is unnecessary since
+  `npm ci` will remove it before installing dependencies. ([#998]https://github.com/heroku/buildpacks-nodejs/pull/998))
+
 ## [3.4.2] - 2025-01-08
 
 - No changes.
--- a/buildpacks/nodejs-npm-install/src/main.rs
+++ b/buildpacks/nodejs-npm-install/src/main.rs
@@ -98,13 +98,6 @@ impl Buildpack for NpmInstallBuildpack {
 
         configure_npm_runtime_env(&context)?;
 
-        let logger =
-            if let Some(prebuilt_modules_warning) = application::warn_prebuilt_modules(app_dir) {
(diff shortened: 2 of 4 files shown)


buildpacks/nodejs-npm-install/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
+
+- Check and warning for `node_modules` directory is unnecessary since
+  `npm ci` will remove it before installing dependencies. ([#998]https://github.com/heroku/buildpacks-nodejs/pull/998))
```
edmorley, line 13: (not blocking) Does `npm ci` warn if it removes `node_modules`? If it doesn't, I wonder if there may end up being a non-zero amount of apps whose builds are slower / more likely to hit git limits due to forgetting to add their `node_modules` to `.gitignore`? Though I suppose if the goal is to encourage optimised apps (rather than as an npm-specific workaround), I guess the check would need to be moved to the engine buildpack anyway, so it displays the warning for yarn/pnpm apps too?

edmorley (approved):

buildpacks/nodejs-npm-install/CHANGELOG.md
colincasey: No, there is no warning. It's implied by the usage of `npm ci` and since the output shows the command when installing, I consider that a loose form of notification.

Could some installs be slower? I hadn't considered the performance impact that would come from the deletion though having your `node_modules` directory checked into git would have a broader impact than just during dependency installation and you'd likely see other negative side-effects. 

Either way, this warning is in the wrong place here but I'm still trying to figure out how to tackle some of these application checks in a way that lets us add Heroku-specific warnings and best practices such as Node version support deprecations (like Python does) without impeding the individual functionality of some of these buildpacks.

edmorley: > Could some installs be slower?

If the user is Git pushing a 1GB `node_modules` directory rather than only the app source, I'm presuming there will be a reasonably performance impact.

PR #1014, opened 2025-02-11
Fix signing for npm inventory mirroring

colincasey:
This was broken by the recent addition of a secondary key to npm's list of public keys. This will unblock jobs like https://github.com/heroku/buildpacks-nodejs/actions/runs/13269889375/job/37046632436

### DIFF
--- a/common/bin/download-verify-npm-package
+++ b/common/bin/download-verify-npm-package
@@ -50,20 +50,27 @@ if [ "$actual_integrity" != "$published_integrity" ]; then
 	exit 1
 fi
 
-# Build the npm signing key from published npm key data. This section assumes there
-# is only one published public key. That was true at the time of
-# writing, but if npmjs.com starts using multiple keys for signatures, this
-# section will need to change. The public key(s) may be obtained from
-# https://registry.npmjs.com/-/npm/v1/keys.
+# Build the npm signing key from published npm key data.
+# The public key(s) may be obtained from https://registry.npmjs.com/-/npm/v1/keys.
+npm_pubkeys=(
+  "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE1Olb3zMAFFxXKHiIkQO5cJ3Yhl5i6UPp+IhuteBJbuHcA5UogKo0EWtlWwW6KSaKoTNEYL7JlCQiVnkhBktUgg=="
+  "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEY6Ya7W++7aUPzvMTrezH6Ycx3c+HOKYCcNGybJZSCJq/fd7Qa8uuAKtdIkUQtQiEKERhAmE5lMMJhP8OkDOa2g=="
+)
+
 echo "Verifying npmjs.com signature..." >&2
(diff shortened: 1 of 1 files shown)


schneems (approved):

common/bin/download-verify-npm-package
```diff
@@ -50,20 +50,27 @@ if [ "$actual_integrity" != "$published_integrity" ]; then
...
 
-# Build the signing message
-echo -n "${package_name}@${package_version}:${published_integrity}" >message.txt
+	# Build the signing message
+	echo -n "${package_name}@${package_version}:${published_integrity}" >message.txt
 
-# Verify the signature
-openssl dgst -sha256 -verify npm-pubkey.pem -signature npm-signature.bin message.txt
+	# Verify the signature
+	if openssl dgst -sha256 -verify npm-pubkey.pem -signature npm-signature.bin message.txt; then
```
edmorley, line 73: @colincasey I might be missing something (this is only via code inspection in the browser diff), but won't this still pass even if none of the signatures match, since the loop will complete and the script end as normal?

schneems: You're right. I modified the script checksums to be always invalid, it reports as a zero exit status

```
$ ./download-verify-npm-package "npm" "11.1.0"
Determining dist url from https://registry.npmjs.com/npm/11.1.0
Downloading npm tarball from https://registry.npmjs.org/npm/-/npm-11.1.0.tgz ...
Saving as ./npm-v11.1.0.tar.gz
Checking npm tarball integrity...
Verifying npmjs.com signature...
- Trying key #1
Could not find private key of public key from npm-pubkey.pem
- Trying key #2
Could not find private key of public key from npm-pubkey.pem
$ echo $?
0
```

schneems: Perhaps not the most elegant way to solve the problem, but this should fix it https://github.com/heroku/buildpacks-nodejs/pull/1015

colincasey: ugh, reason # 999 on why i hate shell scripting and reason # 1 for how code reviews help catch stupid bugs. thanks @schneems and @edmorley 😅

schneems: I checked this one in several ways and missed it. I compared that the lines didn't change and the order of the commands were the same. Though we're well aware that bash has never been a daily driver for me. I am not actually 100% sure why this script exits 0, I'm assuming that either the if or the iterator counts as a final command?

I'm thinking of other similar failures that might be worth checking and/or wondering if there's some ways we could harden our code through checks or structure to guard against them:

While we're checking things, does this pass on the exit of the interior command? I'm assuming it does, but worth getting a second opinion https://github.com/heroku/buildpacks-nodejs/blob/e6fa0d6a0c84ecfbad579d693f30f7b3426432f2/.github/workflows/_mirror-distribution.yml#L58-L62. Would adding a `set` to fail any non-zero exit help harden this in the future?

colincasey: It already does fail which is how I discovered the keys weren't matching. I just got confused by the output from the `openssl` command which prints a failure message and then I mistakenly thought the exit code would be a failure but I never confirmed that.

PR #1021, opened 2025-02-14
Remove metrics

colincasey:
Related to https://github.com/heroku/buildpacks-jvm/pull/679. This may become a dedicated buildpack in the future if metrics still work this way on fir.


buildpacks/nodejs-engine/tests/integration_test.rs
```diff
@@ -1,17 +1,12 @@
 // Required due to: https://github.com/rust-lang/rust/issues/95513
 #![allow(unused_crate_dependencies)]
 
-use libcnb_test::{assert_contains, assert_not_contains, ContainerConfig};
-use std::time::Duration;
+use libcnb_test::assert_contains;
 use test_support::{
     assert_web_response, nodejs_integration_test, nodejs_integration_test_with_config,
-    set_node_engine, wait_for, PORT,
```
edmorley, line 8: (optional) I think the `PORT` const in `test_support` is now unused and could be removed?
https://github.com/heroku/buildpacks-nodejs/blob/077816b7b247add3dec52385659ffbbe300de039/test_support/src/lib.rs#L15

edmorley (approved):

PR #1028, opened 2025-02-21
Improve nodejs-engine errors

colincasey:
This PR adds detailed and consistent error messages for the nodejs-engine buildpack + includes snapshot testing for the output.

### DIFF
--- a/buildpacks/nodejs-engine/src/errors.rs
+++ b/buildpacks/nodejs-engine/src/errors.rs
@@ -103,6 +103,7 @@ fn on_unknown_version_error(version: String) -> ErrorMessage {
         .create()
 }
 
+#[allow(clippy::too_many_lines)]
 fn on_dist_layer_error(error: DistLayerError) -> ErrorMessage {
     match error {
         DistLayerError::TempFile(e) => error_message()
--- a/buildpacks/nodejs-engine/src/errors.rs
+++ b/buildpacks/nodejs-engine/src/errors.rs
@@ -1,0 +1,365 @@
+use crate::install_node::DistLayerError;
+use crate::NodeJsEngineBuildpackError;
+use bullet_stream::style;
+use heroku_nodejs_utils::error_handling::error_message_builder::SetIssuesUrl;
+use heroku_nodejs_utils::error_handling::ErrorType::{Internal, UserFacing};
+use heroku_nodejs_utils::error_handling::{
+    file_value, on_framework_error, ErrorMessage, ErrorMessageBuilder, SuggestRetryBuild,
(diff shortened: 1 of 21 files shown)


buildpacks/nodejs-engine/src/errors.rs
```diff
@@ -0,0 +1,366 @@
...
+    fn test_dist_layer_installation_error() {
+        assert_error_snapshot(DistLayerError::Installation {
+            src_path: "/tmp/path-to-src".into(),
+            dst_path: "/layers/buildpack/path-to-dst".into(),
+            source: create_io_error("unexpected end of file"),
+        });
+    }
+
+    #[test]
+    fn test_dist_layer_checksum_verification_error() {
```
joshwlewis, line 316: I'm not totally seeing the value of tests like these. Really just seems like we're testing the interpolation in the error message and some of the prefix formatting from bullet stream. There's not a lot of code logic in play here to test. These strike me as "I really meant what I said" type tests, where we basically write the same thing twice. I suppose there is some value in that (to prevent accidental changes), but I don't think the juice is worth the squeeze for me. Of course, this is probably a matter of taste, and maybe the overhead is worth it for you.

Where I think these might have more value is with more encompassing logging -- like say making sure we don't write "You specified engines.npm and packageManager in package.json" type messages several times.

joshwlewis (approved):

buildpacks/nodejs-engine/src/errors.rs
colincasey: Think of it less as a unit test and more as a visual regression test. The value proposition is different. Error messages are one of the highest value touch points of our UX. For a lot of users, they don't care about the output until their builds are not working and that's where these messages become critical. So I want the most valuable thing, what an error message looks like to the user, to be as tangible as possible to the following parties:
- Developers
- Code reviewers
- CX

And the best way to make that output tangible is to render it as raw text, as the user would experience it, not as code, which is how we as maintainers experience it. 

A more encompassing logging test might be a benefit but, as you know, integration tests increase the time of test runs. Not to mention that the way that error handling is dealt with in `libcnb.rs` means that we handle all errors that bubble up in one single handler and each error is mapped to a single message so I'm not at all concerned that we might see a "You specified engines.npm and packageManager in package.json" type messages several times. The strategy employed here minimizes the cost of running these tests while maximizing how we understand the experience of the user when they encounter an error.

joshwlewis: Yeah, I get it -- this is definitely just taste, and you are the chef here. =)

PR #1039, opened 2025-02-27
Send all output to stderr

colincasey:
Based on team discussions, it sounds like we want to output our logging to `stderr` which is what this PR does. Also cited in discussion was the following:
- https://unix.stackexchange.com/questions/331611/do-progress-reports-logging-information-belong-on-stderr-or-stdout

### DIFF
--- a/buildpacks/npm/bin/build
+++ b/buildpacks/npm/bin/build
@@ -1,45 +1,47 @@
 #!/usr/bin/env bash
 
-set -e
-set -o pipefail
+{
+  set -e
+  set -o pipefail
 
-bp_dir=$(
-	cd "$(dirname "$0")"/..
-	pwd
-)
-build_dir=$(pwd)
-layers_dir=$1
-platform_dir=$2
+  bp_dir=$(
+    cd "$(dirname "$0")"/..
+    pwd
+  )
+  build_dir=$(pwd)
+  layers_dir=$1
+  platform_dir=$2
 
-# shellcheck source=/dev/null
-source "$bp_dir/lib/build.sh"
-rm -rf "$build_dir/node_modules"
+  # shellcheck source=/dev/null
+  source "$bp_dir/lib/build.sh"
+  rm -rf "$build_dir/node_modules"
 
-#Failures that are caught immediately
-fail_multiple_lockfiles "$build_dir"
+  #Failures that are caught immediately
+  fail_multiple_lockfiles "$build_dir"
 
-clear_cache_on_stack_change "$layers_dir"
+  clear_cache_on_stack_change "$layers_dir"
 
-export_env "$platform_dir/env" "" ""
+  export_env "$platform_dir/env" "" ""
 
-install_or_reuse_toolbox "$layers_dir/toolbox"
-export PATH=$layers_dir/toolbox/bin:$PATH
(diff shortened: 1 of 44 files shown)


buildpacks/npm/bin/build
```diff
@@ -1,45 +1,47 @@
...
 
-run_build "$build_dir"
+  run_build "$build_dir"
 
-write_launch_toml "$build_dir/package.json" "$layers_dir/launch.toml"
+  write_launch_toml "$build_dir/package.json" "$layers_dir/launch.toml"
 
-prune_devdependencies "$build_dir"
+  prune_devdependencies "$build_dir"
+} >&2
```
colincasey, line 47: Simply wrapping the contents of this script in `{ ... } >&2` seems "too easy". Shell scripting has humbled me too many times that this gives me doubt. Is this the best way to redirect all the output from `stdout` to `stderr`?

edmorley: Another way would be to use `exec >&2` on its own line at the top of the file (xref https://unix.stackexchange.com/questions/61931/redirect-all-subsequent-commands-stderr-using-exec) - but either works, and I suppose this Bash buildpack is close to EOL (🤞!), so doesn't matter too much either way :-)

edmorley (commented):
This PR was marked as skip-changelog, but since it does have user facing impact, I think it does need a changelog entry?

colincasey:
Hmm, nothing has changes in the buildpack's operation though. The output destination is different but `pack` displays both the stdout/stderr so wouldn't you consider this more of an internal detail since a user doesn't invoke the buildpack binaries directly? 

Or, by user-facing, do you actually mean platform-facing? And, if that's the case, it looks like the platform spec mandates that implementations output both stdout/stderr when both the [detector](https://github.com/buildpacks/spec/blob/main/platform.md#outputs-1) and [builder](https://github.com/buildpacks/spec/blob/main/platform.md#outputs-4) are invoked, right?

edmorley:
>  The output destination is different but pack displays both the stdout/stderr so wouldn't you consider this more of an internal detail since a user doesn't invoke the buildpack binaries directly?

I'm presuming Pack CLI preserves the stdout vs stderr streams from the buildpack (rather than consolidating them), otherwise we wouldn't see the difference in the `libcnb-test` `pack_stdout` vs `pack_stderr` etc? If so, then this does affect users, since if they did `pack build | grep 'foo'` then it would no longer match etc. We don't make any guarantees about buildpack output, so I don't think it counts as breaking (so eg a minor version bump is fine), but I think anything that could have effect on user should be in the changelog (ie we should err on the side of caution - a changelog entry doesn't cost anything to add :-) ).

joshwlewis (approved):
Looks good, though kind of a bummer we'll need to do this for each of our CNBs -- yet another task to do for 9x languages.

I tend to agree that this probably makes sense as a changelog entry.

edmorley:
(For anyone not on our team following along - part of the context of this, is this Pack and/or lifecycle bug: https://github.com/buildpacks/pack/issues/2330)

colincasey:
I agree with your assessment @edmorley. Added the changelog entries and will ship this as a `minor` change.

PR #1082, opened 2025-04-15
Output snapshots for corepack

colincasey:
This PR replaces manual output assertions with snapshots for better visibility into output changes.

### DIFF
--- a/buildpacks/nodejs-pnpm-install/src/main.rs
+++ b/buildpacks/nodejs-pnpm-install/src/main.rs
@@ -6,6 +6,7 @@
 use crate::configure_pnpm_store_directory::configure_pnpm_store_directory;
 use crate::configure_pnpm_virtual_store_directory::configure_pnpm_virtual_store_directory;
 use bullet_stream::{style, Print};
+use fun_run::CommandWithName;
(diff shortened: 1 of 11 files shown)


buildpacks/nodejs-pnpm-install/src/main.rs
```diff
@@ -142,6 +145,15 @@ impl Buildpack for PnpmInstallBuildpack {
     }
 }
 
+fn disable_pnpm_auto_update_notifier(env: &Env) {
```
edmorley, line 148: I think this change should be in its own PR? This PR has the "skip changelog" label applied, and only mentions updating test snapshots, whereas this addition is a user-facing behaviour change?

colincasey: Sorry, I should have switched this PR to draft after noticing the update notifiers. I wasn't sure if this was what I wanted to do but I was having trouble with getting a filter that I was happy with 😅 

This has been reverted in https://github.com/heroku/buildpacks-nodejs/pull/1082/commits/bb2d25ddeb3e8a2451d0a7d61593a7623bdd64f6

edmorley (approved):

PR #1092, opened 2025-04-24
Add retries to binary downloads

### LINKED ISSUE #868: Add retries to Node.js binary download
To reduce the chance of intermittent failures like:

```
[Heroku Node.js Engine Buildpack]

[Checking Node.js version]
Detected Node.js version range: >=20.0.0 <21.0.0-0
Resolved Node.js version: 20.15.0

[Installing Node.js distribution]
Downloading Node.js 20.15.0 (linux-amd64) from https://nodejs.org/download/release/v20.15.0/node-v20.15.0-linux-x64.tar.gz

[Error: Node.js engine distribution error]
Couldn't download Node.js distribution: I/O error while downloading file: response body closed before all bytes were read
```

As seen in:
https://github.com/heroku/cnb-builder-images/actions/runs/9699157499/job/26767524369?pr=541#step:6:27
https://github.com/heroku/cnb-builder-images/actions/runs/9699157499/job/26767523738?pr=541#step:6:28

colincasey:
Swapped out the usage of `download_file` from `libherokubuildpack` which does not currently perform retries for a combination of `reqwest` + retry middleware. Except for a couple of error messages, this is mostly a drop-in replacement.

One drawback here is that an async runtime (`tokio`) is required to make this work. These buildpacks were written without async in mind, so the `download_file_sync` function added in this PR hides most of this away. This isn't the most optimal way to handle this, but due to where the downloads happen, it shouldn't be an issue.

Closes #868


common/nodejs-utils/src/download_file.rs
```diff
@@ -0,0 +1,106 @@
...
+        .map_err(|e| DownloadError::WriteFile(to_file.to_path_buf(), from_url.to_string(), e))?;
+
+    Ok(())
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum DownloadError {
+    #[error("Request to {0} failed\nError: {1}")]
+    Request(String, reqwest_middleware::Error),
+    #[error("Reading response from request to {0} failed\nError: {1}")]
```
schneems, line 100:
```suggestion
    #[error("Reading response from request to `{0}` failed\nError: {1}")]
```

common/nodejs-utils/src/download_file.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    io::copy(&mut response_data.as_slice(), &mut output_file)
+        .map_err(|e| DownloadError::WriteFile(to_file.to_path_buf(), from_url.to_string(), e))?;
+
+    Ok(())
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum DownloadError {
+    #[error("Request to {0} failed\nError: {1}")]
```
schneems, line 98:
```suggestion
    #[error("Request to `{0}` failed\nError: {1}")]
```

common/nodejs-utils/src/download_file.rs
```diff
@@ -0,0 +1,106 @@
...
+use std::fs::File;
+use std::io;
+use std::path::{Path, PathBuf};
+use std::time::Duration;
+
+const DEFAULT_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+
+const DEFAULT_READ_TIMEOUT: Duration = Duration::from_secs(10);
+
+const DEFAULT_RETRIES: u32 = 5;
```
schneems, line 14: Comment: As an FYI I dug into the underlying code and "5 max retries" means "5 max sleeps" but actually 6 requests. Not a problem, just thought it was worth sharing.

common/nodejs-utils/src/download_file.rs
```diff
@@ -0,0 +1,106 @@
...
+    let client = ClientBuilder::new(
+        reqwest::ClientBuilder::new()
+            .use_rustls_tls()
+            .connect_timeout(connect_timeout)
+            .read_timeout(read_timeout)
+            .build()
+            .expect("Should be able to construct the HTTP client"),
+    )
+    .with(RetryTransientMiddleware::new_with_policy(
+        ExponentialBackoff::builder().build_with_max_retries(max_retries),
```
schneems, line 61: I'm new to this API so I played with the library:

```rust
use retry_policies::{RetryDecision, RetryPolicy, policies::ExponentialBackoff};
use std::time::{Duration, SystemTime};

fn main() {
    let backoff = ExponentialBackoff::builder().build_with_max_retries(5);
    for n_past_retries in 0..=6 {
        let now = SystemTime::now();
        let decision = backoff.should_retry(now, n_past_retries);

        eprintln!("Prior: {n_past_retries}");
        match decision {
            RetryDecision::Retry { execute_after } => {
                let countdown = execute_after.duration_since(now).unwrap();
                eprintln!("Retry in {countdown:?}");
            }
            RetryDecision::DoNotRetry => {
                eprintln!("Do not retry");
            }
        }
    }
}
```

```
Prior: 0
Retry in 909.835371ms
Prior: 1
Retry in 973.948531ms
Prior: 2
Retry in 2.794485778s
Prior: 3
Retry in 1.076410877s
Prior: 4
Retry in 12.043339666s
Prior: 5
Do not retry
```

That's is approx 17.4 seconds (`0.909 + 0.973 + 2.794 + 1.07 + 12` ). I'm feeling that is an unexpectedly long amount of maximum time. I think we could tune this down to a max of 1\~1.5 second per retry (by default). That would bring it to a maximum of 5~7.5 seconds. 

Do you have any extra information guiding the values here?

common/nodejs-utils/src/download_file.rs
```diff
@@ -0,0 +1,106 @@
+use reqwest::IntoUrl;
+use reqwest_middleware::ClientBuilder;
+use reqwest_retry::policies::ExponentialBackoff;
+use reqwest_retry::RetryTransientMiddleware;
+use std::fs::File;
+use std::io;
+use std::path::{Path, PathBuf};
+use std::time::Duration;
+
+const DEFAULT_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
```
schneems, line 10: In https://github.com/heroku/heroku-buildpack-ruby/pull/175 a customer asked to increase the default from 3 to 30 seconds. I ended up making it configurable. I think someone running `pack build` in Australia on a tetherd 3g mobile should be slow, but should be possible. 

I think we need to expose a way to allow customers to increase this value if they need, either now or add an issue to our backlog. My suggestion would be something in `project.toml` like `com.heroku.connect.timeout = 30` (spitballing).

common/nodejs-utils/src/download_file.rs
schneems, line 100:
```suggestion
    #[error("Reading response from request to `{0}` failed\nError: {1}")]
```

buildpacks/nodejs-yarn/src/snapshots/errors___yarn_cli_layer_download_error.snap
```diff
@@ -2,7 +2,8 @@
 source: buildpacks/nodejs-yarn/src/errors.rs
 ---
 - Debug Info:
-  - HTTP error while downloading file: http: invalid format
+  - Request to https://test/error failed
+    Error: error sending request for url (https://test/error)
```
schneems, line 6: Do we have any control over the middleware error formatting? I like that the URL is wrapped in character's but it's a slightly different convention from our other locations (backticks). Not a deal breaker, but if it isn't too time-consuming it would look more consistent

buildpacks/nodejs-engine/Cargo.toml
```diff
@@ -26,6 +26,8 @@ toml = "0.8"
 insta = "1"
 libcnb-test = "=0.28.1"
 regex = "1"
+reqwest = "0.12"
+reqwest-middleware = "0.4"
 serde_json = "1"
 test_support.workspace = true
-ureq = "3"
+tokio = "1"
```
schneems, line 33: Suggestion: There is a lot of repetition in the version of all these cargo.toml files. I suggest we move them to the workspace top level `Cargo.toml` and then these would become `tokio.workspace = true }`. It will make future version bumps easier and their diffs smaller, and prevent a scenario where common and a buildpack accidentally get out of sync.

This isn't a blocker, we could move that to the backlog to do an audit update the rest at once.

common/nodejs-utils/src/download_file.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    let client = ClientBuilder::new(
+        reqwest::ClientBuilder::new()
+            .use_rustls_tls()
+            .connect_timeout(connect_timeout)
+            .read_timeout(read_timeout)
+            .build()
+            .expect("Should be able to construct the HTTP client"),
+    )
+    .with(RetryTransientMiddleware::new_with_policy(
```
schneems, line 60: Suggestion:  Let's tell the user when we're retrying if we can. This is instrumented with the tokio tracing crate. We should enable it to collect traces, possibly we can hook into it to emit or change output. That might require investigation and could go into the backlog.

```
                    #[cfg(feature = "tracing")]
                    log_retry!(
                        self.retry_log_level,
                        "Retry attempt #{}. Sleeping {:?} before the next attempt",
                        n_past_retries,
                        duration
                    );
```

common/nodejs-utils/src/download_file.rs
schneems, line 98: I want to know if it's failing due to accidental starting or ending trailing whitespace, also brings it closer to bullet stream style:

```suggestion
    #[error("Request to `{0}` failed\nError: {1}")]
```

schneems (changes requested):
I left some error formatting suggestions. I also left some "I think we should X" suggestions that we could either do now while we're at it, or put in the backlog.

buildpacks/nodejs-engine/Cargo.toml
colincasey: We did have that at one point until @edmorley found that using workspace dependencies wouldn't work right with automated Dependabot updates. I believe this is still an issue as https://github.com/dependabot/dependabot-core/issues/7896 is still open.

common/nodejs-utils/src/download_file.rs
colincasey: So the tracing part is separate from the logging. There's no direct mechanism in the retry middleware to expose this, nor do we necessarily want the low-level tracing provided. The solution would be a custom middleware like what's shown [here](https://github.com/TrueLayer/reqwest-middleware/blob/main/reqwest-retry/src/retryable_strategy.rs#L27-L39). How would you expect those retries to look in the context of `bullet_stream` output?

buildpacks/nodejs-yarn/src/snapshots/errors___yarn_cli_layer_download_error.snap
colincasey: No control over the error formatting, no. You'll get whatever the library provides for messages which is why this is in the debug info section.

common/nodejs-utils/src/download_file.rs
colincasey: It could be configurable. I don't think it should block this PR though. We should discuss this more broadly with the team, especially if it's going to be a generalized configuration like `com.heroku.connect.timeout`.

common/nodejs-utils/src/download_file.rs
colincasey: @schneems this is a rough translation of the current retry numbers used by the classic Node.js buildpack [here](https://github.com/heroku/heroku-buildpack-nodejs/blob/main/lib/binaries.sh#L45). The timings are exponential so you should expect to see the numbers increase with each subsequent retry +/- whatever random jitter is applied. 

I don't see any problems with the `17.4s` your test arrived at since that is a "worst case" scenario. I also don't want to stress a system that's having stability issues by flooding it with too many requests and the exponential strategy is friendlier to that.

common/nodejs-utils/src/download_file.rs
schneems: I've got a way to cancel and restart timers here https://github.com/heroku-buildpacks/bullet_stream/pull/10. I've not implemented it anywhere yet but I think Rune has in the dotnet buildpack. 

Alternatively: Just print dots, and at the end, emit a summary or information. Spitballing:

```
- NPM
  - Downloading ................... (Done 17s)
  - Warnings during download:
  
      Retrying 1 of 3 (Connect timeout retrying in 1s)
      Retrying 2 of 3 (Connect timeout retrying in 2s)
      Retrying 3 of 3 (Connect timeout retrying in 4s)

```

This exact output can't be done directly today, but I've wanted to add a `print::pre()` function to print raw indented output in the style of a run command. It would be pretty easy to add I think, it might make a good first PR.

I think that the important part is that we surface behavior so if there's chronic or pathological issues, they're brought to our attention. In the case I gave above about "tethering from Australia" we want to expose retries and why they're happening to hint that they should perhaps adjust the connect timeout.

usiegj00:
Thanks guys. I’m impressed by the amount of work that went into adding retry logic for downloads, but it strikes me as a lot of effort. The further we stray from the expressive, iterable shell scripts that originally inspired buildpacks, the less maintainable things feel--I'm starting to feel like I’ve been handed three seashells and told it’s obvious. Was the full rewrite of the buildpacks and their supporting infrastructure in Rust driven solely by Windows support, or were there other goals frozen in with it? Either way, congratulations on completing this--it was breaking buildpacks that worked on the non-CNB stacks.

colincasey:
@usiegj00 The choice of Rust over shell scripts might seem like extra effort but this transition has paid off in other ways for buildpack maintainers such as myself. Better tooling and type guarantees, more reliable runtime behaviors, robust error-handling, multi-platform support, improved code sharing mechanisms, etc. Like anything, it's a trade-off. 

For example, Rust makes the failure cases for actions very explicit. It's more work to handle those failure cases but it also allows a buildpack maintainer to provide very precise and prescriptive messages if an error occurs. To me, improving that UX and delivering more clarity to the end-user is a positive outcome for that extra effort.

usiegj00:
@colincasey Thanks again for getting this over the line. I’m totally onboard with Rust’s predictability and the problems it solves for shell, but in practice the trade-offs still feel significant:

| Dimension               | Shell one-liner (`curl --retry`) | Rust implementation (`reqwest` + `tokio`) |
|-------------------------|----------------------------------|-------------------------------------------|
| **LoC / surface area**  | ~1 line                          | +1 683 / −123 across 29 files             |
| **Dependencies**        | `curl`                           | `tokio`, `reqwest`, `backoff`, …          |
| **Retry semantics**     | Fixed back-off, no jitter        | Configurable policy, richer errors        |
| **Portability**         | POSIX; Windows needs workarounds | Cross-compiles; Windows “just works”      |
| **Debug & override**    | `bash -x`, env overrides         | Requires Rust toolchain, structured logs  |
| **Interactive debugging** | Run/inspect each command live in the terminal | Edit → compile → run cycle                |
| **Contributor barrier** | Shell is ubiquitous              | Rust knowledge now required               |
| **Security surface**    | Shell + external binary          | Memory-safe code, larger crate tree       |
| **Delivery timeline**   | Feature already existed          | ~10 months (Issue #868 → merge)           |

### Open questions
1. **When do we reach for Rust?**  
   What decides which future plumbing stays in shell versus jumps to Rust?
2. **Parity vs. divergence**  
   If more buildpacks migrate, how do we guarantee that the Rust versions reproduce today’s behavior byte-for-byte so the theoretical gains show up in practice?

Looking forward to the conversation—and appreciate all the heavy lifting!

edmorley:
@usiegj00 One of the other considerations (in our internal RFC back when we were making our decision some time ago) was that the upstream CNB spec is more complicated than the classic buildpack API, and as such more involved to implement:
https://github.com/buildpacks/spec/blob/main/buildpack.md

For example, creating/parsing TOML files (and in particular nested toml tables) is something that would be quite painful using a pure Bash solution - both to implement and to reason about correctness wise (due to lack of type system). 

Having now spent a while working with CNBs in Rust, I'm very much glad that we didn't use Bash for our CNBs - and every time I have to go back to Bash to work on the classic buildpack it feels like a step backwards. Bash definitely has it's place for shorter sized scripts, but for anything larger/more complicated it's extremely painful to work with even when using tooling like Shellcheck to try and avoid some of the footguns (and I say this as someone who's pretty good at writing Bash and that has also recently written a lot of Bash while modernising the classic Python buildpack).

PR #1093, opened 2025-04-25
Use `bullet_stream` global functions instead of state-based ones

colincasey:
This PR might looks like a lot but the individual commits are logically separated. What's happening here is, all of the state-based output functions from `bullet_stream` have been replaced with their `bullet_stream::global::print` counterparts. This removes a lot of the clutter added to weave the state-based logger throughout each buildpack's codebase. 

The end result is the same build output (which you can see in the snapshots), except for:
- the buildpack headers which are now prefixed with `##` instead of `#` in `bullet_stream@0.8.0`
- the custom command names used for executing pnpm and Yarn build scripts have been dropped in favor of the default command name provided by `fun_run`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -822,7 +822,7 @@ dependencies = [
 name = "heroku-pnpm-engine-buildpack"
 version = "0.0.0"
 dependencies = [
- "bullet_stream 0.7.0",
+ "bullet_stream 0.8.0",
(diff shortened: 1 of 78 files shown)


schneems (approved):
Breaking up the commits make this easy to review. Even easier with the snap files to see what changed. Looks great, thanks for updating the header style!

PR #1113, opened 2025-05-23
Unify bullet_stream and fun_run versions with workspace

schneems:
The `bullet_stream` library has a feature that exposes interfaces using a fun_run defined trait. This feature can be used without requiring fun_run, but if a developer wants more control of their errors or for manipulating how the commands are printed then they need to also import the `fun_run` library. This gets tricky in Rust because (unlike Ruby) the same crate can be imported using multiple different versions depending on how the dependencies resolve. Frustratingly, if you try to use the exact same struct with the exact same fields and behavior in a signature that was defined using a different version, then it fails to compile and often without a terribly helpful message. 

Even when all crates can resolve to the same version of the same library, sometimes cargo will prefer to use different versions based on the state of the lockfile. This PR unifies all `bullet_stream` and `fun_run` versions to rely on the same workspace defined version and updates them to the latest released versions.


colincasey:
@schneems this can be closed. #1114 took care of resolving the version discrepancy from the workspace lockfile. Also, workspace dependencies still aren't handled properly by dependabot or we would be using them already.

schneems:
You're welcome to close anything you won't merge. Ideally, with a message about why (like this one above).

> workspace dependencies still aren't handled properly by dependabot or we would be using them already.

Huh, TIL (or today I remembered, not sure which) https://github.com/dependabot/dependabot-core/issues/7896. It's my preference to use workspaces and unify all sub-crates. The uv folks do the same https://corrode.dev/podcast/s04e03-astral/.

This behavior sounds fine for the most part: https://github.com/dependabot/dependabot-core/issues/7896#issuecomment-2714402481. It seems most crates are very conservative with their versioning and semver compatibility. And ones that rev big numbers need bigger work (like ureq 2->3). 

I'm now wondering if we could hack a github action to "toml edit" our way to a once-a-month cheap dependabot workspace updater.

edmorley:
> I'm now wondering if we could hack a github action to "toml edit" our way to a once-a-month cheap dependabot workspace updater.

@schneems I think a first steps before spending engineering time on this creating our own Dependabot replacement, would be to:
(a) update the linked GitHub issue with one that doesn't use grouping (since when writing the original issue I thought it was related to grouping, but it seems it's not, so the testcase can be simplified)
(b) opening a GitHub support ticket (given we have an enterprise account) pointing at the issue and asking for someone to take a look at it.

PR #1118, opened 2025-06-11
Refactor download file

colincasey:
This PR is in preparation for changes to how npm and Yarn binaries are resolved, downloaded, and installed. For this, I've split the original `download_file` function into two distinct parts:
- `get(url)` which handles making retryable requests
- `response.download_to_file(path)` which streams a response directly to the filesystem

This also removes some of the message customization the previous `download_file` builder supported as I wasn't happy with how that design was working. As these utility functions are meant to encapsulate common buildpack actions, I think this acceptable to drop as the current output, while not configurable, will at least be consistent.

Here is what the previous output was:

```sh
- Downloading Node.js `22.16.0 (<arch>)` from https://nodejs.org/download/release/v24.1.0/node-v24.1.0-linux-x64.tar.gz ... (timeout)
- Retry attempt 1 of 2 ... (0.1s)
```

And here is what this PR changes it to:

```sh
- GET https://nodejs.org/download/release/v24.1.0/node-v24.1.0-linux-x64.tar.gz ... (timeout)
- Retry attempt 1 of 2 ... (0.1s)
- Downloading ... (12s)
```

> [!NOTE]
> The original output contained a bug where only the request/response timing was captured, not the complete request/response + read body.


schneems (approved):
LGTM. I also played with this with some badly behaved servers and it seems to do what I expect https://gist.github.com/schneems/8e9c6685cca2a5ac6f5760af310a6a65.

PR #1121, opened 2025-06-13
Replace npm inventory with packument

colincasey:
This PR replaces the inventory-based npm version resolution strategy with an equivalent method that uses the same process npm itself uses.

For example, if you run `npm info npm@latest`, a request will be made to https://registry.npmjs.com/npm to fetch what npm calls a `Packument`. The `Packument` is in JSON format and contains metadata about all published versions of a package under the `versions` field. This `versions` field is essentially a richer version of what our inventory file contained as each entry has the package version, distribution url, checksums, etc.

While it also could have been possible to replace npm version resolution with a call to `npm info npm@{requested_version}`, the advantage of requesting the `Packument` directly is that the buildpack can cache this information for use in subsequent builds. The cached `Packument` will be invalidated whenever a new version of `npm` is published.


buildpacks/nodejs-npm-engine/src/main.rs
```diff
@@ -73,16 +73,27 @@ impl Buildpack for NpmEngineBuildpack {
...
+        print::bullet("Determining npm package information");
+        let npm_package_packument =
+            resolve_requested_npm_packument(&context, &requested_npm_version)?;
 
         print::bullet("Installing npm");
-        let npm_release = resolve_requested_npm_version(&requested_npm_version, &inventory)?;
-        install_npm(&context, &npm_release, &node_version)?;
-        log_installed_npm_version(&env)?;
+        if existing_npm_version == npm_package_packument.version {
+            print::sub_bullet("Requested npm version is already installed");
```
schneems, line 87: This output doesn't show up in any of the snapshots. Does it need to be exercised under test or is this branch not actually called? The `install_npm` also has cached logic. If possible I would prefer a single point of truth so we don't have to remember to update two locations if the logic changes.

common/nodejs-utils/src/npmjs_org.rs
```diff
@@ -24,3 +34,135 @@ pub(crate) fn list_releases(package: &str) -> anyhow::Result<Vec<NpmRelease>> {
         .map_err(|e| anyhow!("Couldn't serialize npmjs registry release list for {package}: {e}"))
         .map(|rel| rel.versions.into_values().collect())
 }
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct PackumentMetadata {
+    etag: Option<String>,
+    last_modified: Option<String>,
+}
```
schneems, line 42: Can we add a cache invalidation serialization test? An example is https://github.com/heroku/buildpacks-ruby/blob/fc2a18615eb27f53d0ea5f35a94f5f63aa000032/buildpacks/ruby/src/layers/ruby_install_layer.rs#L208-L215. 

If someone modifies the metadata, they won't get feedback that their actions will cause cache invalidations. For example:

```diff
$ git diff
diff --git a/common/nodejs-utils/src/npmjs_org.rs b/common/nodejs-utils/src/npmjs_org.rs
index ff5e68c..039e289 100644
--- a/common/nodejs-utils/src/npmjs_org.rs
+++ b/common/nodejs-utils/src/npmjs_org.rs
@@ -39,6 +39,7 @@ pub(crate) fn list_releases(package: &str) -> anyhow::Result<Vec<NpmRelease>> {
 struct PackumentMetadata {
     etag: Option<String>,
     last_modified: Option<String>,
+    yolo: String,
 }

 pub fn packument_layer<B, E>(
@@ -116,6 +117,7 @@ where
         packument_layer.write_metadata(PackumentMetadata {
             etag,
             last_modified,
+            yolo: "yolo".to_string(),
         })?;
     } else if packument_response.status() == StatusCode::NOT_MODIFIED {
         print::sub_bullet(format!("Using cached packument for {package_name}"));
```

buildpacks/nodejs-corepack/tests/snapshots/corepack_npm_8.snap
```diff
@@ -7,7 +7,7 @@ source: test_support/src/lib.rs
   - Node.js version not specified, using `22.x`
   - Resolved Node.js version: `22.16.0`
 - Installing Node.js distribution
-  - GET https://nodejs.org/download/release/v22.16.0/node-v22.16.0-<arch>.tar.gz ... (<time_elapsed>)
+  - GET https://nodejs.org/download/release/v22.16.0/node-v22.16.0-<arch>.tar.gz ... (OK)
```
schneems, line 10: To consider for later: There's a "url" style that wraps in backticks and colorized. The output is coming from a `Url` type so that would catch any stray extra spaces and error on them, so it's not critical.

schneems (approved):
I'm out next week. Approved with some enhancement suggestions and some questions to validate.

buildpacks/nodejs-corepack/tests/snapshots/corepack_npm_10.snap
```diff
@@ -7,7 +7,7 @@ source: test_support/src/lib.rs
   - Node.js version not specified, using `22.x`
   - Resolved Node.js version: `22.16.0`
 - Installing Node.js distribution
-  - GET https://nodejs.org/download/release/v22.16.0/node-v22.16.0-<arch>.tar.gz ... (<time_elapsed>)
+  - GET https://nodejs.org/download/release/v22.16.0/node-v22.16.0-<arch>.tar.gz ... (OK)
```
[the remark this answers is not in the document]
colincasey: Yes, though it's a bit more nuanced than that since the output line above doesn't fully encapsulate the complete timing of the HTTP request. The output / timing is actually split into two parts - receiving headers + reading the body. E.g.;

- HTTP request start
  - Output → `GET {url}` 
- Response 
  - Receive headers  
    - Output → `GET {url} ... (OK)`
  - Output → `Downloading`
  - Read body
    - Output → `Downloading ........................... (1m 41s)`
- HTTP request complete 

And because the response headers are typically received quickly, the timing output didn't seem as interesting to include there. I think it's much more informative to show the status code instead and leave the timing to the reading of the response body as that's where the majority of the request time is spent

Issue #1142, opened 2025-07-12
YARN2_SKIP_PRUNING no longer respected in Yarn 4


jaisal1024:
Our builds starting failing yesterday as YARN2_SKIP_PRUNING=true as an environment variable is no longer respected. We need to flip the prune step off but cannot figure out how to. 

I think this is related to release 4.0.0 where pruning was re-introduced. Its a great change but we need a way to opt out

To reproduce run:

 pack build <app>--builder heroku/builder:24 --env NODE_OPTIONS=--max_old_space_size=4096 ---env 'YARN2_SKIP_PRUNING=true'"

Rohithgilla12:
@colincasey I am trying to build a nestjs application.

I even tried doing this build command by adding the version to the buildpack, but the built image still fails. 
```
 pack build ${{ env.IMAGE_PATH_AR }} \
            --path ./backend \
            --tag ${{ env.IMAGE_PATH_AR }}:latest \
            --builder heroku/builder:22 \
            --buildpack heroku/nodejs@3.6.11 \
```

Everything was working till a few days back,  I will try to attach logs

colincasey:
Yes, downgrading the version will work. 

Version `4.0.0` of the Node.js CNB does expose configuration to opt-out of dev dependency pruning but currently only through buildplan metadata so later buildpacks can disable this feature if necessary (e.g.; #1141). I also plan to expose the ability to opt-out of dev dependency pruning through user configuration in `project.toml` but that's not supported yet.

PR #1146, opened 2025-07-16
Expose dev dependency pruning as a project.toml config 

colincasey:
Currently, the Node.js CNBs support a similar configuration via the buildpack plan from contributing buildpacks and there's no reason why this also shouldn't be configurable by end users.

#### Usage Example in `project.toml`

```toml
# _.schema-version is required for the project descriptor
[_]
schema-version = "0.2"

[com.heroku.buildpacks.nodejs]
actions.prune_dev_dependencies = false
```

> [!WARNING]
> Both the `project.toml` and buildplan configurations are experimental features and subject to change in the future as the designs for CNB configuration are refined.

Closes https://github.com/heroku/buildpacks-nodejs/issues/1142


common/nodejs-utils/src/config.rs
```diff
@@ -0,0 +1,338 @@
...
+use std::path::Path;
+use std::str::FromStr;
+use toml_edit::TableLike;
+
+const CONFIG_NAMESPACE: [&str; 5] = ["com", "heroku", "buildpacks", "nodejs", "actions"];
+const PRUNE_DEV_DEPENDENCY_CONFIG: &str = "prune_dev_dependencies";
+
+pub fn read_prune_dev_dependencies_from_project_toml(
+    project_toml_path: &Path,
+) -> Result<Option<bool>, ConfigError> {
```
schneems, line 16: [Observation, no action needed] When I see a (non env-var) config, I'll ask if there's something that it can be beyond a boolean. Like if we wanted to have different states beyond on/off for pruning dev dependencies. Like "after_install" or "later" or "force" or something. In this case we don't yet have the abilities to do those things, but if there's a strong chance we might in the future, it's easier to bake it in now, so we don't have to do a config change later. 

I know we're already somewhat down this road with build plan consistency, so I don't think it merits a change now, but while we're looking at it, it's good to consider possible desired use cases (or explicitly state certain usecases are not considered valid or supported).

schneems, line 16: Suggestion (not blocking): I see an opportunity to reduce/remove the need to create tempfiles for tests, if the interface is split in two:

- `fn toml_edit_from_path(path: &Path) -> Result<Option<toml_edit::DocumentMut>, ConfigError> {`
- `fn dev_dependencies_config(doc: &toml_edit::DocumentMut) -> Result<Option<bool>, ConfigError> {`

You would still need tempfiles to exercise `toml_edit_from_path`, but the testing of the `dev_dependencies_config` could take a direct input that can be constructed without file creation.

The type signature return of the first (`toml_edit_from_path`) forces you to handle the "file does not exist" externally.

common/nodejs-utils/src/config.rs
```diff
@@ -0,0 +1,338 @@
...
+    fn from(value: ConfigError) -> Self {
+        let project_toml = style::value("project.toml");
+        let toml_spec_url = style::url("https://toml.io/en/v1.0.0");
+        let config_table_key = style::value(format!("[{}]", CONFIG_NAMESPACE.to_vec().join(".")));
+        let config_key = style::value(PRUNE_DEV_DEPENDENCY_CONFIG);
+
+        match value {
+            ConfigError::CheckExists(e) => error_message()
+                .error_type(ErrorType::Internal)
+                .header("Failed to check if project.toml exists")
```
schneems, line 78: The subject order implies that the buildpack forgot "failed to check" rather than it checked and the check failed.

```suggestion
                .header("Failed project.toml existence check")
```

common/nodejs-utils/src/config.rs
```diff
@@ -0,0 +1,338 @@
...
+                    This buildpack will read from {project_toml} if the file is present in the root of \
+                    the application but an unexpected I/O error occurred during this operation.
+                "})
+                .debug_info(e.to_string()),
+
+            ConfigError::ParseProjectToml(e) => error_message()
+                .error_type(UserFacing(SuggestRetryBuild::Yes, SuggestSubmitIssue::No))
+                .header("Failed to parse project.toml")
+                .body(formatdoc! { "
+                    This buildpack reads configuration from {project_toml} to complete the build but \
```
schneems, line 98: Grammar, missing comma:

```suggestion
                    This buildpack reads configuration from {project_toml} to complete the build, but \
```

common/nodejs-utils/src/config.rs
```diff
@@ -0,0 +1,338 @@
...
+                    - Ensure the file follows the TOML format described at {toml_spec_url}
+                "})
+                .debug_info(e.to_string()),
+
+            ConfigError::ExpectedTomlTable => error_message()
+                .error_type(UserFacing(SuggestRetryBuild::Yes, SuggestSubmitIssue::No))
+                .header("Error parsing project.toml with invalid key")
+                .body(formatdoc! { "
+                    This buildpack reads the configuration from {project_toml} to complete \
+                    the build but the configuration for the key {config_table_key} isn't the \
```
schneems, line 111:
```suggestion
                    the build, but the configuration for the key {config_table_key} isn't the \
```

common/nodejs-utils/src/config.rs
```diff
@@ -0,0 +1,338 @@
...
+                    {toml_spec_url}
+                " })
+                .maybe_debug_info(None::<String>),
+
+            ConfigError::WrongType => error_message()
+                .error_type(UserFacing(SuggestRetryBuild::Yes, SuggestSubmitIssue::No))
+                .header("Error parsing project.toml with invalid configuration type")
+                .body(formatdoc! { "
+                    This buildpack reads the configuration from {project_toml} to complete \
+                    the build but the configuration value for {config_key} in the section \
```
schneems, line 125:
```suggestion
                    the build, but the configuration value for {config_key} in the section \
```

buildpacks/nodejs-npm-install/tests/snapshots/test_skip_build_scripts_from_buildplan.snap
```diff
@@ -36,4 +36,7 @@ source: test_support/src/lib.rs
   - Skipping as pruning was disabled by a participating buildpack
 - Configuring default processes
   - Skipping default web process (no start script defined)
+
+! Experimental feature used - buildpack plan configuration (this feature may change unexpectedly in the future)
```
schneems, line 40: Suggestion (not blocking): I think we should tell people specifics of where the warning is coming from. Either list the key and file here or earlier. 

```
! Warning: Experimental configuration `com.heroku.buildpacks.nodejs.actions` found in `project.toml`.
! This feature may change unexpectedly in the future
```

buildpacks/nodejs-npm-install/tests/snapshots/test_prune_dev_dependencies_config.snap
```diff
@@ -0,0 +1,40 @@
...
+
+      added <NUMBER> packages, and audited <NUMBER> packages in <time_elapsed>
+
+      found 0 vulnerabilities
+
+  - Done (<time_elapsed>)
+- Running scripts
+  - No build scripts found
+- Pruning dev dependencies
+  - Skipping as pruning was disabled in project.toml
```
schneems, line 34: Suggestion (not blocking): I think we should tell people specifics of where the warning is coming from. Either list the key and file here or earlier so we don't have to make them guess (even if it's fairly obvious now, lets set a positive pattern so if we add a lot more keys in the future they won't have to guess).

schneems (approved):
Left some optional suggestions here and there. Overall looks good. Need to re-apply snapshot updates to current tests.

common/nodejs-utils/src/config.rs
colincasey: Fixed in bc36e48a

common/nodejs-utils/src/config.rs
colincasey: Fixed in bc36e48a

common/nodejs-utils/src/config.rs
colincasey: Fixed in bc36e48a

buildpacks/nodejs-npm-install/tests/snapshots/test_skip_build_scripts_from_buildplan.snap
colincasey: Fixed in 415f79c2

schneems (approved):
Thanks for the update. Looks good.

PR #1169, opened 2025-09-02
Refactor individual buildpacks into one singular buildpack

colincasey:
## Context

Early on in the development of the Node.js CNBs, the idea was that, by leveraging certain features of the CNB spec (e.g; provides/requires, composite buildpacks, etc.), we would benefit from improved composability which would allow us to support more use cases, add new Node.js tooling faster, and so on. Now, several years into this approach, I no longer think it's the best choice for how this repository should be organized.

The major downside to using the micro-buildpack approach is that we want to provide a simple and consistent user experience when using our CNBs on the Heroku platform. Individual CNBs produce more duplication and effort to maintain this consistency at a technical level (testing, metrics, output, error messaging, configuration, documentation). It buys us very little since we want users to always be using our `heroku/nodejs` composite CNB and not the individual CNBs so we can guarantee expected behavior.

Instead of having multiple buildpacks that each provide a small piece of functionality, a single buildpack that provides equivalent functionality for all Node.js / JavaScript related tools will require less overhead and remove impediments that make providing a transition path from our classic Node.js buildpack harder.

## Contents

This PR is the initial phase of refactoring work required to replace our multiple Node.js CNBs:

- Component Buildpacks
  - `heroku/nodejs-corepack`
  - `heroku/nodejs-engine`
  - `heroku/nodejs-npm-engine`
  - `heroku/nodejs-npm-install`
  - `heroku/nodejs-pnpm-engine`
  - `heroku/nodejs-pnpm-install`
  - `heroku/nodejs-yarn`
- Composite Buildpacks
  - `heroku/nodejs` 

With a single, component buildpack `heroku/nodejs` that encompasses all of the above. 

This may seem like a lot of changes but it mostly includes mechanical changes like moving files to new locations, deleting unnecessary files, and fixing module imports. This can be seen in the updated snapshots from integration tests which are almost unchanged except for:
- individual buildpack titles and timings are no longer printed as this has been replaced with a single `heroku/nodejs` title and timing at the start and end of the build
- pack output for included buildpack ids, caching layers, etc. now only references the single `heroku/nodejs` buildpack

There most significant code changes are around each of the `main.rs` files from the individual CNBs which needed their `detect`, `build`, and `on_error` functions to be extracted from their (no longer needed) buildpack structs. Once separated from the buildpack structs, they can be folded into the top-level buildpack in a way that [recreates the former composite `heroku/nodejs` buildpack's ordered groups](https://github.com/heroku/buildpacks-nodejs/blob/main/meta-buildpacks/nodejs/buildpack.toml).

## Notable Changes

There are two notable changes here that deviate from this refactor being a true reproduction of the previous individual CNBs:

1. In the top-level `detect`, we always pass and provide `heroku/nodejs` for later buildpacks that may might need to optionally require Node.js functionality. If a `package.json`, `index.js`, or `server.js` file is detected then we also require `heroku/nodejs` to support standalone Node.js applications. This means that we no longer provide `node`, `npm`, `pnpm`, `yarn`, `node_modules`, and `node_build_scripts`. 
2. The `npm-install` buildpack used to provide a sanity check and error when multiple lockfiles were found during the build phase. This had to be dropped and will be reintroduced in later changes in a way that allows this check to be performed for all package managers before the Node.js runtime is downloaded.

### Deprecation Plan

The existing buildpacks found in the `./buildpacks` directory have been retained for the initial v5 release. These buildpacks provide the following error message during `detect` if they are used:

```
! Usage of `{buildpack_id}` is deprecated and will no longer be supported beyond v4.1.4.
!
! Equivalent functionality is now provided by the `heroku/nodejs` buildpack:
! - Buildpacks authors that previously required `{buildpack_id}` should now require `heroku/nodejs` instead.
! - Users with a `project.toml` file that lists `{buildpack_id}` should now use `heroku/nodejs` instead.
!
! If you have any questions, please file an issue at https://github.com/heroku/buildpacks-nodejs/issues/new.
```

This will only be seen if a buildpack author or user tries to use the `5.0.0` version of any of the deprecated buildpacks. These buildpacks will be permanently removed after this initial release.

For our builder images, any buildpacks that reference <sup>([1](https://github.com/heroku/cnb-builder-images/blob/main/builder-24/builder.toml#L73-L96), [2](https://github.com/heroku/cnb-builder-images/blob/main/builder-22/builder.toml#L73-L96))</sup> these deprecated buildpacks will need to be manually edited to ensure they are not updated as part of our automated release process until the affected buildpacks can be updated.

## Future Changes

After this massive restructuring it will be possible to make smaller, more targeted changes to clean up the code and remove existing duplication.


edmorley (approved):
🎉

PR #1179, opened 2025-09-15
Add retries to CI integration tests

colincasey:
- Uses [cargo-nextest](https://nexte.st/) as the integration test runner since it has retry support.
- Also uses SHA for `actions/checkout` version.


Malax (commented):
I'd love for this PR to have an explanation why we need this :)

colincasey:
Ah, double-entry problem! The rationale was explained in the linked work item. I've copied it over here for visibility @Malax  😅

PR #1249, opened 2025-12-05
Switched downloader from parallel to sequential actions

colincasey:
The download implementation used in this buildpack was attempting to provide an optimized process that would stream a download directly to a target destination while also handling checksum validation and gzip-compressed tar extraction. This relied heavily on async capabilities to support this pipeline of actions.

While this optimized process appeared to function correctly, over time, two patterns began to emerge that hinted this might not be true:
- Internal instrumentation showed intermittent checksum failures when no request or extraction errors were present which would indicate that the checksum digest wasn't writing all the bytes from the stream.
- End-to-end testing on amd-based architecture would intermittently segfault at the extraction stage.

Neither of these cases was reproducible locally but both hint at some timing-dependent error at either the checksum stream or gzip-decompression → tar extraction stream.

Given this, I've decided to switch to a less-optimal but simpler download process with three sequential stages:
1. request → temporary file
2. temporary file → checksum validation
3. temporary file → extraction

My theory is that the removal of the asynchronous work will improve stability and eliminate the intermittent errors observed. If not, and these intermittent errors persist, at least it will be clearer which step in the process is the culprit.

[W-20389228](https://gus.lightning.force.com/a07EE00002QPcchYAD)


edmorley (approved):
(Rubber stamp to unblock figuring out the segfaults on Direwolf)

PR #1401, opened 2026-06-25
Classify npm install-block failures into actionable build errors

colincasey:
[npm v12 will refuse install scripts, git dependencies, and remote-URL dependencies by default](https://github.blog/changelog/2026-06-09-upcoming-breaking-changes-for-npm-v12/) unless an app explicitly opts in. Today these failures all surface as the buildpack's single generic "Failed to install Node modules" error, which tells a user nothing about why the install was rejected or how to unblock it — they're left to guess from the raw npm output.

This routes the three install-blocking npm error codes (`ESTRICTALLOWSCRIPTS`, `EALLOWGIT`, `EALLOWREMOTE`) into specific build errors that name what npm refused and point to the relevant remediation (approving scripts, or permitting git/remote dependencies in `.npmrc`), rather than nudging users to disable npm's protections buildpack-wide. Any unrecognized install failure still falls through to the existing generic message unchanged.

W-23160935

### DIFF
--- a/src/package_managers/npm.rs
+++ b/src/package_managers/npm.rs
@@ -2,7 +2,7 @@ use crate::cleanup::{CleanupTask, NodeGypArtifactLocation};
 use crate::utils::build_env::node_gyp_env;
 use crate::utils::error_handling::ErrorType::Internal;
 use crate::utils::error_handling::{
-    ErrorMessage, ErrorType, SuggestRetryBuild, SuggestSubmitIssue, error_message,
+    ErrorMessage, ErrorType, SuggestRetryBuild, SuggestSubmitIssue, error_codes, error_message,
 };
 use crate::utils::npm_registry;
 use crate::{BuildpackBuildContext, BuildpackResult};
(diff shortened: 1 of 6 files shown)


src/package_managers/npm.rs
```diff
@@ -219,6 +299,24 @@ mod tests {
...
+                &format!("printf '%s' {} >&2; exit 1", shell_quote(stderr)),
+            ])
+            .named(name)
+            .named_output()
+            .unwrap_err()
+    }
+
+    fn shell_quote(s: &str) -> String {
+        format!("'{}'", s.replace('\'', "'\\''"))
+    }
```
edmorley, line 318: This seems extremely hacky - it's running a subprocess to output text to work around the suboptimal API design of `fun_run`, and then performing DIY shell escaping (which could be extremely fragile). We should either (a) find another way to do this here, (b) fix `fun_run`, (c) stop using `fun_run` (it has quite  a few other issues and is drastically over-engineered IMO).

src/utils/error_handling.rs
```diff
@@ -141,6 +141,49 @@ pub(crate) enum SuggestSubmitIssue {
...
+            }
+
+            fn from_code(code: &str) -> Option<$name> {
+                $name::iter().find(|e| e.code() == code)
+            }
+        }
+    };
+}
+
+pub(crate) use error_codes;
```
edmorley, line 176: I'm not sure this macro is the best approach:
1. It adds complexity (see below for non-macro version - much fewer lines/cognitive burden) for not much benefit (Clippy + tests will catch most cases drift between the `NpmError` variants and the `from_code()` impl).
2. It breaks if any of the future NPM emitted error codes aren't also valid Rust identifiers.
3. It is less flexible for cases that might come up in the future (eg scenarios like checking if two strings appear in the stderr, to avoid false positive matches, or using a regex to match fuzzy cases etc)

A non-macro version could look something like:

```rust
  pub(crate) enum NpmError {
      StrictAllowScripts,
      AllowGit,
      AllowRemote,
  }

  impl NpmError {
      fn from_code(code: &str) -> Option<NpmError> {
          match code {
              "ESTRICTALLOWSCRIPTS" => Some(NpmError::StrictAllowScripts),
              "EALLOWGIT"           => Some(NpmError::AllowGit),
              "EALLOWREMOTE"        => Some(NpmError::AllowRemote),
              _ => None,
          }
      }
  }
```

colincasey: I did have something that looked roughly like your suggested non-macro version initially but I deliberately changed it for the macro version which I think will add value here (maybe it just seems premature at this stage?). The suggestion you provide does have the ability to drift between a new `NpmError` variant added and the `from_code` implementation which you do mention can be minimized in (1) but, in the macro version, there is no potential for drift at all.

As for (2), ~~I'm not sure what you mean since the macro makes you provide the error code and Rust identifier. It's not derived from the error code so the identifier can't be invalid.~~ (just realized you meant the `$code:ident` part for the error code). Fair point, I can easily address that by changing the macro to:
```
error_codes!(NpmError {
    StrictAllowScripts => "ESTRICTALLOWSCRIPTS",
    AllowGit           => "E-ALLOW-GIT",   // hyphens, dots, anything now fine
    AllowRemote        => "EALLOWREMOTE",
});
```

And this design is not meant to cover (3). It's only meant to cover error cases where npm emits a single, well-documented error code. From my investigation into how npm uses these codes, I don't see this being an issue right now. If it does turn out to be insufficient in the future then the design can change then.

src/package_managers/npm.rs
colincasey: I agree, it's hacky. For test-only code, I'm not terribly concerned. Lot's of Rust errors are inconvenient to simulate 😅  

I will make a PR to `fun_run` to get this improved.

src/utils/error_handling.rs
colincasey: @edmorley Would you be okay with a slimmed-down version that drops `iter()`/`code()` and just generates `from_code` as a direct match — keeping the single source of truth so added codes can't drift into a missing arm?

src/package_managers/npm.rs
edmorley: @colincasey Ah I missed it was test only - in which case I'm fine with it for now :-)

src/utils/error_handling.rs
edmorley: > It's only meant to cover error cases where npm emits a single, well-documented error code.

Ah I was thinking of cases where npm (or tooling it calls) emits some other string that could be multiple words and not a status code.

If this is just for the fairly narrow npm error code case, then agree (3) won't be needed, and (2) is less likely to occur - so I'd be fine with punting dealing (2) until/if it happens, and if you really do prefer the macro, then that's ok.

edmorley (approved):

src/package_managers/npm.rs
colincasey: All good. I filed https://github.com/schneems/fun_run/pull/22 which will make it easier to validate command failures because I do like the conveniences that `fun_run` provides.

src/utils/error_handling.rs
colincasey: I've slimmed down the macro to only the essentials. In this case, I think it'll show value as I expand on the npm error codes and hopefully this pattern will translate nicely to Yarn/pnpm too.

Issue #1175, opened 2025-09-09
Support service bindings for additional config files (e.g.; .npmrc)


sermler:
Hello,

i recently tried to update to v5.0.1 but it seems, that the "heroku-preinstall" script is no longer executed before running pnpm install.
I used it like mentioned in https://github.com/heroku/heroku-buildpack-nodejs/issues/363#issuecomment-273278852 to configure a private registry.

Is it possible to get it back? 
Or alternatively the servicebindings https://github.com/buildpacks/spec/blob/main/extensions/bindings.md could be used for this.(That's the way the paketo buildpacks handle additional config files).

I can try to provide a PR for this, if this would be an option.

PR #1416, opened 2026-07-16
Feature/add service binding support

SimonErm:
Implements https://github.com/heroku/buildpacks-nodejs/issues/1175 by adding a basic service-binding implementation and copying binded configs to target paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,54 @@ The following scripts from `package.json` will be executed by the configured pac
  
 If any of the above scripts are not defined in `package.json` they will be skipped.
 
+### Private npm Registry via Service Bindings
+
+Configure access to private npm registries using [Kubernetes Service Bindings](https://github.com/k8s-service-bindings/spec)(The recommended format: <https://github.com/buildpacks/spec/blob/main/extensions/bindings.md>).
+
+Supported Bindings:
+
+|type|secret|
+|----|-------|
+|npm|.npmrc|
+|yarn|.yarnrc|
+|pnpm|auth.ini|
+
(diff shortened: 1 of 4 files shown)


colincasey:
@SimonErm do you have a real use case that requires adding service binding support to this CNB?

SimonErm:
@colincasey yes, i described it at <https://github.com/heroku/buildpacks-nodejs/issues/1175#issuecomment-3310723478>

Am 22. Juli 2026 19:59:31 MESZ schrieb Colin Casey ***@***.***>:

colincasey:
@SimonErm, while I appreciate this contribution, I'm also hesitant about accepting it for the following reasons:
- Even though the CNB Go frameworks ([libcnb](https://github.com/buildpacks/libcnb), [libpak](https://github.com/paketo-buildpacks/libpak), [packit](https://github.com/paketo-buildpacks/packit)) support the [service binding spec](https://servicebinding.io/), its actual usage is concentrated to a subset of Paketo CNBs.
- Heroku has no mechanism to leverage service bindings on the platform and is the primary target for this CNB.

I do like how service bindings solve the problem of providing sensitive auth details to tools like the Node.js package managers but I worry about maintaining code that would never be used on Heroku and (possibly) rarely be used outside of Heroku. If we were to go forward and add service binding support, it would have to be clearly marked as "experimental" and subject to unexpected changes or removal and limited support.

I would also recommend paring this feature down to a smaller initial scope. For example, only adding type-based service binding support for a single package manager like npm. That would make it easier to review and evaluate.

PR #1438, opened 2026-09-01
Add pnpm 12 support

colincasey:
pnpm 12 is a ground-up rewrite that changes how the CLI is distributed: rather than a single JavaScript package, the native executable now ships as a separate platform-specific package (`@pnpm/exe.<target>`) pulled in as an optional dependency. The buildpack's pnpm installer assumed the older single-package layout, so apps pinning pnpm 12 couldn't install it.

These changes allow the buildpack to adapt to the new pnpm 12 distribution model and install the native binary matching the build's target architecture.

W-24059616

### DIFF
--- a/src/package_managers/pnpm.rs
+++ b/src/package_managers/pnpm.rs
@@ -4,7 +4,11 @@ use crate::utils::build_env::node_gyp_env;
 use crate::utils::error_handling::{
     ErrorMessage, ErrorType, SuggestRetryBuild, SuggestSubmitIssue, error_message, file_value,
 };
-use crate::utils::npm_registry::{PackagePackument, packument_layer, resolve_package_packument};
+use crate::utils::npm_registry::{
+    InstallPackageLayerError, PackagePackument, apply_package_layer_env, cached_package_layer,
+    exclude_package_docs, extract_package_tarball, install_package_layer_metadata,
+    link_package_bins, packument_layer, resolve_package_packument,
+};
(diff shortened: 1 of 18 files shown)


src/package_managers/pnpm.rs
```diff
@@ -42,16 +46,150 @@ pub(crate) fn install_pnpm(
...
+        |path| path != Path::new("pnpm"),
+    )
+    .map_err(|e| {
+        InstallPackageLayerError::Download(Box::new(native_binary_packument.clone()), e)
+    })?;
+
+    let native_pnpm_binary = scratch_dir.join("pnpm");
+    let original_pnpm_binary = pnpm_layer_dir.join("pnpm");
+    fs::rename(&native_pnpm_binary, &original_pnpm_binary)
+        .map_err(|e| create_pnpm_native_binary_install_error(&original_pnpm_binary, &e))?;
```
schneems, line 159: It's not guaranteed that the problem is coming this path `original_pnpm_binary` and not `native_pnpm_binary`. We're only showing one in the resulting error we bubble up.

std::fs::rename can fail in either the from or two path https://doc.rust-lang.org/std/fs/fn.rename.html

- If to does not exist, from can be anything.
- On Unix, when from is a directory and to exists, to must be an empty directory.
- On Unix, when from is not a directory and to exists, to may not be a directory.

I would first suggest using fs-err which captures **both** paths by default and you don't need to worry about where the error mode comes from. Then you just capture the std::io::Error and they're already in there.

schneems (approved):

